PR #64, opened 2019-02-27
Watch every resource managed by the operator

martonsereg:
If something changes - like a deployment is mistakenly deleted - the operator will reconcile the state.


waynz0r:
Two things is currently under development for the PR.

- the related object reconcile triggering is going to be configurable
- fix #73

PR #86, opened 2019-03-05
fix port name in gateway deployment

waynz0r:


pkg/resources/gateways/deployment.go
```diff
@@ -156,7 +156,7 @@ func (r *Reconciler) ports(gw string) []apiv1.ContainerPort {
 			{ContainerPort: 443, Protocol: apiv1.ProtocolTCP, Name: "https"},
 			{ContainerPort: 31400, Protocol: apiv1.ProtocolTCP, Name: "tcp"},
 			{ContainerPort: 15029, Protocol: apiv1.ProtocolTCP, Name: "https-kiali"},
-			{ContainerPort: 15030, Protocol: apiv1.ProtocolTCP, Name: "https-prometheus"},
+			{ContainerPort: 15030, Protocol: apiv1.ProtocolTCP, Name: "https-prom"},
```
Laci21: Why did you change this name? Shouldn't you change it [here](https://github.com/banzaicloud/istio-operator/blob/istio11-fix-gateway-deployment/pkg/resources/gateways/service.go#L48) as well then?

waynz0r: I've got the following error, `could not reconcile istio: failed to reconcile resource: updating resource failed: Deployment.apps \"istio-ingressgateway\" is invalid: spec.template.spec.containers[0].ports[4].name: Invalid value: \"https-prometheus\": must be no more than 15 characters`

Laci21: Ah, I see. Thanks for the commit!

waynz0r: For some reason that limit is different at the service definition, but i changed the name there as well.

PR #136, opened 2019-03-19
CNI plugin support

### LINKED ISSUE #135: CNI plugin implementation

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #135 

### Checklist

- [x] Implementation tested (GKE, EKS)


config/samples/istio_v1beta1_istio_cni_gke.yaml
```diff
@@ -0,0 +1,69 @@
...
+    replicaCount: 1
+    minReplicas: 1
+    maxReplicas: 5
+  sidecarInjector:
+    image: "docker.io/istio/sidecar_injector:1.1.0-rc.4"
+    replicaCount: 1
+    rewriteAppHTTPProbe: true
+    initCNIConfiguration:
+      enabled: true
+      binDir: /home/kubernetes/bin
```
Laci21: The only difference between `istio_v1beta1_istio_cni.yaml` and `istio_v1beta1_istio_cni_gke.yaml` is this line: `binDir: /home/kubernetes/bin`. The rest of the yaml is the same all the way. I'm wondering whether we should use some kind of templating for these samples to get rid of duplicated lines and make the samples easier to maintain. For example, do you think using `kustomize` would be too much here?

pkg/apis/istio/v1beta1/defaults.go
```diff
@@ -30,13 +30,17 @@ const (
 	defaultSDSImage               = defaultImageHub + "/" + "node-agent-k8s" + ":" + defaultImageVersion
 	defaultProxyImage             = defaultImageHub + "/" + "proxyv2" + ":" + defaultImageVersion
 	defaultProxyInitImage         = defaultImageHub + "/" + "proxy_init" + ":" + defaultImageVersion
+	defaultInitCNIImage           = defaultImageHub + "/install-cni:master-latest-daily"
```
Laci21: Nitpicking:
```suggestion
	defaultInitCNIImage           = defaultImageHub + "/" + "install-cni" + ":" + "master-latest-daily"
```

config/samples/istio_v1beta1_istio_cni_gke.yaml
Laci21: The only difference between `istio_v1beta1_istio_cni.yaml` and `istio_v1beta1_istio_cni_gke.yaml` is this line: `binDir: /home/kubernetes/bin`. The rest is the same all the way. I'm wondering whether we could use some kind of templating to get rid of the duplicated lines and make the samples easier to maintain. For example, do you think using `kustomize` would be too much here?

pkg/apis/istio/v1beta1/defaults.go
Laci21: Nitpicking:
```suggestion
	defaultInitCNIImage           = defaultImageHub + "/" + "install-cni" + ":" + "master-latest-daily"
```

PR #137, opened 2019-03-19
Prepare release with Istio 1.1.0 support

waynz0r:


pkg/apis/istio/v1beta1/defaults.go
```diff
@@ -19,8 +19,8 @@ package v1beta1
 import "fmt"
 
 const (
-	defaultImageHub               = "gcr.io/istio-release"
-	defaultImageVersion           = "release-1.1-latest-daily"
+	defaultImageHub               = "docker.io"
```
martonsereg: I think it should be `docker.io/istio`

waynz0r: Nice catch, thanks @martonsereg!

PR #163, opened 2019-04-09
Use code-generator to generate clientset, listers, informers

Laci21:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

- Use [code-generator](https://github.com/kubernetes/code-generator) to generate clientset, listers, informers.
- Update circleCI config to use workflow.


Laci21:
> are these files generated or written manually?
> My guess is that it's not generated, but I'd rather go with a generated approach, so we could have informers and listers as well, and it's also less error prone. We could use https://github.com/kubernetes/code-generator to do that.

Good idea, updated the code to use code-generator.

waynz0r:
thanks @Laci21! 👍

PR #218, opened 2019-05-26
Fix immediate resource re-queuing

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

Disables immediate resource re-queuing where it should wait for a bit.

- [x] Implementation tested

### DIFF
--- a/pkg/controller/istio/istio_controller.go
+++ b/pkg/controller/istio/istio_controller.go
@@ -238,7 +238,7 @@ func (r *ReconcileConfig) reconcile(logger logr.Logger, config *istiov1beta1.Ist
 	if config.Status.Status == istiov1beta1.Reconciling {
 		logger.Info("cannot trigger reconcile while already reconciling")
 		return reconcile.Result{
-			Requeue:      true,
+			Requeue:      false,
 			RequeueAfter: time.Duration(30) * time.Second,
 		}, nil
 	}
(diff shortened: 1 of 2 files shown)


pkg/controller/istio/istio_controller.go
```diff
@@ -238,7 +238,7 @@ func (r *ReconcileConfig) reconcile(logger logr.Logger, config *istiov1beta1.Ist
 	if config.Status.Status == istiov1beta1.Reconciling {
 		logger.Info("cannot trigger reconcile while already reconciling")
 		return reconcile.Result{
-			Requeue:      true,
+			Requeue:      false,
```
Laci21: If this is explicitly set to `true` then requeuing happens immediately instead of after the given timeframe? Because this is what I understand from your PR description but this is not what I'd understand from the description [here](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/reconcile/reconcile.go#L31). Also, if I check the [code](https://github.com/kubernetes-sigs/controller-runtime/blob/b9f30df3f3a81bad9902b8bec5d92622a363bd4f/pkg/internal/controller/controller.go#L216), I would think that the `Requeue` field's value should not have any effect if `RequeueAfter` is set to a non zero value.

waynz0r: It seems that you're right :)

My initial assumption about how this works was right after all, I close this PR.

PR #225, opened 2019-05-30
Kustomize v2 compatibility

mousavian:
| Q               | A
| --------------- | ---
| Bug fix?        | no|yes
| New feature?    | no|yes
| API breaks?     | no|yes
| Deprecations?   | no|yes
| Related tickets | fixes #X, partially #Y, mentioned in #Z

### What's in this PR?
Moving default `kustomization.yaml` configuration from `config/default` to `config`.
And taking `manager_image_patch.yaml` file out of the base and put it under `config/custom` directory.

### Why?
So that it'll be compatible with kustomize v2 as in version 2 it's not possible to reference resources outside of root dir.
This will allow us to have our own kustomization.yaml file with a reference to yours as a base without cloning the repo like below:
```yaml
bases:
  - github.com/banzaicloud/istio-operator/config/base?ref=release-1.1
```

### Additional context
- https://github.com/kubernetes-sigs/kustomize/issues/776
- https://community-banzaicloud.slack.com/archives/CGKQF5HUP/p1559292931039900

### Checklist
- [x] Implementation tested


config/kustomization.yaml
```diff
@@ -18,15 +18,15 @@ namePrefix: istio-operator-
...
   # which protects your /metrics endpoint.
-- ../rbac/auth_proxy_service.yaml
-- ../rbac/auth_proxy_role.yaml
-- ../rbac/auth_proxy_role_binding.yaml
+- rbac/auth_proxy_service.yaml
+- rbac/auth_proxy_role.yaml
+- rbac/auth_proxy_role_binding.yaml
 
 patches:
 - manager_image_patch.yaml
```
Laci21: You have moved the `kustomization.yaml` file from `config/default` to `config` (along with `manager_auth_proxy_patch.yaml` and `manager_prometheus_metrics_patch.yaml`). There is however, this `manager_image_patch.yaml` file which should be located under `config/default` according to the [Makefile](https://github.com/banzaicloud/istio-operator/blob/3804b23719199ada4a7e8f78484c04186c9d8a08/Makefile#L91). (This file is added to [.gitignore](https://github.com/banzaicloud/istio-operator/blob/3804b23719199ada4a7e8f78484c04186c9d8a08/.gitignore#L32) so that it is not overwritten by developers who use different images for development.)

In this line, simply `manager_image_patch.yaml` is referenced from the `config` directory, but to align with the Makefile it should be searched under `config/default`.

```suggestion
- default/manager_image_patch.yaml
```

(Alternatively, this line could stay like this and the line in the [Makefile](https://github.com/banzaicloud/istio-operator/blob/3804b23719199ada4a7e8f78484c04186c9d8a08/Makefile#L91) could be changed to `./scripts/image_patch.sh ./config/manager_image_patch.yaml ${IMG}`, but I'd rather keep this file separately under `config/default`.)

mousavian: Good catch @Laci21 ,
I forgot to fix this one. Thanks for reviewing.

mousavian: Do you reckon I should move `manager_auth_proxy_patch.yaml` and `manager_prometheus_metrics_patch.yaml` back to `default` directory as well? 🤔 

I'm not really across the history behind this image_patching so I've asked question in slack...

Makefile
```diff
@@ -88,12 +88,13 @@ install: manifests
...
 	kubectl apply -f config/manager/namespace.yaml
-	./scripts/image_patch.sh ./config/default/manager_image_patch.yaml ${IMG}
+	./scripts/image_patch.sh ./config/custom/manager_image_patch.yaml ${IMG}
 	bin/kustomize build $(KUSTOMIZE_BASE) | kubectl apply -f -
 
 # Generate manifests e.g. CRD, RBAC etc.
 manifests:
-	go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go all
+	go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go rbac --output-dir config/base/rbac
+	go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go crd --output-dir config/base/crds
```
mousavian: I wasn’t able to use https://github.com/kubernetes-sigs/controller-tools/blob/master/cmd/controller-gen/main.go#L207-L208
as it belongs to v0.2.0 with some breaking changes. 
Instead I used `--output-dir` to achieve the same goal; but I had to use separate commands instead of `all`

mousavian:
I would should suggest we keep `config/default/kustomization.yaml` (instead of moving it to `config/custom`) with below content for backward compatibility for those who have integration with this file, so we don't have to bump major version:
```
bases:
- ../base

patches:
- manager_image_patch.yaml
```

mousavian:
Also what do you think about creating two different kustomization configs under `config/custom` , each one uses base config with separate patches:

```yaml
# config/custom/manager_auth_proxy_enabled.yaml
bases:
- ../base

patches:
- manager_image_patch.yaml
- manager_auth_proxy_patch.yaml
```


```yaml
# config/custom/manager_prometheus_scraping_enabled.yaml
bases:
- ../base

patches:
- manager_image_patch.yaml
- manager_prometheus_metrics_patch.yaml
```

Makefile
Laci21: I see, this looks fine to me, thanks!

Laci21:
> I would should suggest we keep `config/default/kustomization.yaml` (instead of moving it to `config/custom`) with below content for backward compatibility for those who have integration with this file, so we don't have to bump major version:
> 
> ```yaml
> # config/default/kustomization.yaml (for backward compatibility)
> bases:
> - ../base
> 
> patches:
> - manager_image_patch.yaml
> - manager_auth_proxy_patch.yaml
> ```

I like your idea thinking about backward compatibility, but still I wouldn't keep such a file because of the following reasons:
-  We are striving for backward compatibility for sure, but since we haven't released an actual major version yet (0.1.16 is the latest image right now), I think it is acceptable to do such a change.
- More importantly, exactly why you've created this PR anyone would have had trouble when trying to use the operator as a `kustomize` base. So I think it's fair to assume that no one actually used it like that before.
- Also, I would simply avoid having another `kustomization.yaml` if possible.

@martonsereg, @waynz0r what do you guys think?

Laci21:
> Also what do you think about creating two different kustomization configs under `config/custom` , each one uses base config with separate patches:
> 
> ```yaml
> # config/custom/auth_proxy_enabled/kustomization.yaml
> bases:
> - ../../base
> 
> resources:
> - rbac/auth_proxy_service.yaml
> - rbac/auth_proxy_role.yaml
> - rbac/auth_proxy_role_binding.yaml
> 
> patches:
> - manager_auth_proxy_patch.yaml
> ```
> 
> ```yaml
> # config/custom/prometheus_scraping_enabled/kustomization.yaml
> bases:
> - ../../base
> 
> patches:
> - manager_prometheus_metrics_patch.yaml
> ```

I like the idea, I think this would be nice. Would you like to add it to this PR?

Laci21:
@mousavian, I've added a few commits if you don't mind, updating a few paths which was needed for example for the circle tests to pass.
Please, let us know if you would like to add the two different kustomization configs, otherwise, I hope we can merge this soon.

mousavian:
Thank you so much @Laci21 , I've moved auth_proxy,prometheus, and custom_image configs into `overlays` to follow the [standard naming](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/#bases-and-overlays).

Laci21:
@mousavian, I think there were some issues with the naming of a few resources (the main issue was with the `kustomization.yaml` in the `auth-proxy-enabled` dir, because it not only has `patches`, but has new `resources` defined as well and the `namePrefix` defined in the base yaml was not used for the objects defined under `resources`). The point is that I hope I've solved these issues with the above commits :)

I've also added a `README.md` to describe how to use the `kustomization.yaml` files from a different project.

Let me know if the changes made makes sense to you or if there is anything else you would like to add to this PR.

mousavian:
I've noticed that CRDs now unnecessarily have separate kustomization file (`config/base/kustomization.yaml`) whereas they could be part of base, so we don't have to define two bases or run build twice. Specially because kustomize takes care of ordering. But it should be fine for now. 
Thanks.👍

Laci21:
> I've noticed that CRDs now unnecessarily have separate kustomization file (`config/kustomization.yaml`) whereas they could be part of base, so we don't have to define two bases or run build twice. Specially because [kustomize takes care of ordering](https://github.com/kubernetes-sigs/kustomize/blob/af57fc3ecedea20b617062ae12e4bbba895e5418/pkg/gvk/gvk.go#L95). But it should be fine for now.
> Thanks.👍

I haven't fully described why I've separated the crd and namespace creation into another `kustomization.yaml`, so let me do it now. In your proposed solution, where the crds and namespace are built from the same `kustomization.yaml`, I think there was an issue with the `namePrefix` being set to `istio-operator-`. With that being set, `kustomize` didn't try to create a namespace called `istio-system` rather one called `istio-operator-istio-system`. (The same issue was present with the CRDs naming.) This was the reason why these reosurces were originally simply created with `kubectl apply -f` in `make deploy`.

However, with this PR, we are aiming to build a solution where you don't necessarily need to clone this repo to use it, rather you can use it as a `kustomize` base. For that, if there wouldn't be this additional `config/kustomization.yaml` file which I created, then even if you use a base e.g. `config/overlays/auth-proxy-enabled` then to make it work you would need to first create the crds and the namespace for yourself. That is why I came up with this new `config/kustomization.yaml`, so that you can set this (`config/kustomization.yaml`) as a base as well to your `kustomization.yaml` and no need to apply anything by hand.

Let me know, if this makes sense to you. Also, if you have a better idea with one `kustomization.yaml` file while also solving the naming issue I mentioned, then I totally up for it. I could not come up with one yesterday.

mousavian:
Hey @Laci21 , thanks for clarifying.
Yeah it's totally fine for now. We can always improve later.
Thank you again for your consideration and effort to make this happen.

Laci21:
Thank you very much for your contribution @mousavian!

mousavian:
@Laci21 Thank you all for reviewing and helping to get this merged.
It is valuable to us and I really appreciate it.
Cheers.

Laci21:
I'm glad to hear that @mousavian!
Good luck using it and feel free to let us know in case you have any issues or feature requests!

PR #227, opened 2019-06-03
Fix CoreDNS configmap reconciliation

### LINKED ISSUE #226: Corrupting CoreDNS ConfigMap on reconsolidating istiocoredns component
**Describe the bug**
Upon creating new `istio.istio.banzaicloud.io` resource, when `istiocoredns` component is not enabled, `istiocoredns` still gets reconsolidated and causing corrupt coredns ConfigMap with wrong proxy (missing destination): 
```
global:53 {
  errors
  cache 30
  proxy .
}
```

And after few seconds, when coredns pods are reloaded due to config change, they will throw error and stuck in `CrashLoopBackOff` state:
```
$ kubectl log -n kube-system coredns-76b76ccc9-zs8cv
plugin/proxy: /etc/coredns/Corefile:20 - Error during parsing: Wrong argument count or unexpected line ending after '.'
```

This is due to having empty `clusterIP` when `desiredState != k8sutil.DesiredStatePresent`
https://github.com/banzaicloud/istio-operator/blob/ae4867d279/pkg/resources/istiocoredns/coredns.go#L52-L73

**Steps to reproduce the issue:**
`kubectl apply -f config/samples/istio_v1beta1_istio_minimal.yaml`

**Expected behavior**
If I'm not mistaken, `istiocoredns` reconsolidation should only happen when `istiocoredns` is enabled.

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #226 

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/resources/istiocoredns/coredns.go
+++ b/pkg/resources/istiocoredns/coredns.go
@@ -124,8 +124,10 @@ func (r *Reconciler) updateCorefile(corefile []byte, config caddyfile.EncodedSer
 		}
 	}
 
-	if remove && pos > 0 {
-		corefileJSON = append(corefileJSON[:pos], corefileJSON[pos+1:]...)
+	if remove {
+		if pos > 0 {
+			corefileJSON = append(corefileJSON[:pos], corefileJSON[pos+1:]...)
+		}
 	} else {
 		if pos > 0 {
 			corefileJSON[pos] = config


mousavian:
Thanks for quick fix!

Issue #233, opened 2019-06-13
request new feature: configure PodSecurityPolicy in Istio CR


JoshuaAndrew:
**Is your feature request related to a problem? Please describe.**

if  I enable podsecuritypolicy in kube-apiserver, you would find the problem below, there is no 
available pod running, so we should add some configuration in Istio CRD to configure podsecuritypolicy name.
```
[root@k8s-master-1 istio]# kubectl get deploy -nistio-system
NAME                     DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
istio-citadel            1         0         0            0           6m35s
istio-egressgateway      1         0         0            0           6m33s
istio-galley             1         0         0            0           6m35s
istio-ingressgateway     1         0         0            0           6m34s
istio-pilot              1         0         0            0           6m34s
istio-policy             1         0         0            0           6m32s
istio-sidecar-injector   1         0         0            0           6m22s
istio-telemetry          1         0         0            0           6m32s
```

waynz0r:
Hi @JoshuaAndrew,

Thank you for your suggestion!

We are gonna look into it and figure out an easy way to configure PSP in the CR for the Istio components.

JoshuaAndrew:
@waynz0r    wonderful !

JoshuaAndrew:
@waynz0r  Is there any progress?

waynz0r:
Hi @JoshuaAndrew,

we had several internal discussions about how to solve this problem. We think it is out of scope of the operator itself to specify and set pod security policies. 

We are going to add a basic PSP to the helm chart and to the kustomize flow to support PSP enabled clusters.

waynz0r:
Hi @JoshuaAndrew,

we just released a new helm chart version (0.0.14). You can use `rbac.psp.enabled=true` to enable psp   resource deployment for the operator and the Istio components.

Please let us know how it works for you!

JoshuaAndrew:
@waynz0r  Nice cool !

Morriz:
well, thanks for attempt #1, but it is fully permissive...not recommended:
```
spec:
  fsGroup:
    rule: RunAsAny
  runAsUser:
    rule: RunAsAny
  seLinux:
    rule: RunAsAny
  supplementalGroups:
    rule: RunAsAny
  volumes:
  - secret
  - configMap
  - emptyDir
```

I suggest creating more finegrained psps for the components. Which components need access to what?

waynz0r:
[...]

PR #250, opened 2019-07-01
Specify istio-init user explicitely (#249)

### LINKED ISSUE #249: Istio sidecar init-container requires to run as `root`
**Describe the bug**

This bug is already describe here: https://github.com/istio/istio/issues/10358

Is there a way to configure the istio-init container with root access while the pod is configure with less requirements ?

allamand:
Signed-off-by: sebastien allamand <sebastien.allamand@orange.com>

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #249

### What's in this PR?

This PR goal it to configure explicitely the istio-init container with root userID so that it can do iptables.

### Why?

Without this, the container inherit the Pods capabilities and if it is configure to be nonRoot user, the istio-init container will fail

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/resources/sidecarinjector/configmap.go
+++ b/pkg/resources/sidecarinjector/configmap.go
@@ -84,6 +84,7 @@ func (r *Reconciler) proxyInitContainer() string {
   imagePullPolicy: ` + string(r.Config.Spec.ImagePullPolicy) + `
 ` + r.getFormattedResources(r.Config.Spec.SidecarInjector.Init.Resources, 2) + `
   securityContext:
+    runAsUser: 0
     capabilities:
       add:
       - NET_ADMIN


allamand:
@Laci21, I've again problem if my Pod as RunAsNonRoot policy :

```
 Error: container's runAsUser breaks non-root policy
```

In this thread  https://github.com/istio/istio/issues/6806 they say it's better to uses https://github.com/istio/cni. 

Does the operator is compatible with cni ?

thanks!

Laci21:
> @Laci21, I've again problem if my Pod has RunAsNonRoot policy :
> 
> ```
>  Error: container's runAsUser breaks non-root policy
> ```
> 
> In this thread [istio/istio#6806](https://github.com/istio/istio/issues/6806) they say it's better to uses https://github.com/istio/cni.
> 
> Does the operator is compatible with cni ?
> 
> thanks!

Hi @allamand,

thanks for trying out the operator!

If your pod has `RunAsNonRoot` policy set explicitly then I believe you won't be able to use the `istio-init` container as it specifically requires `root` user to set up some `iptables` rules.

However, the CNI plugin was added to Istio exactly because of the this security concern. With CNI, there is no need for the elevated privileges as it performs traffic redirection in the pod’s network setup phase. The operator does support this scenario, for more information refer to [this blog post](https://banzaicloud.com/blog/istio-cni-plugin/) and to [this sample CR](https://github.com/banzaicloud/istio-operator/blob/release-1.1/config/samples/istio_v1beta1_istio_cni.yaml) (or [this](https://github.com/banzaicloud/istio-operator/blob/release-1.1/config/samples/istio_v1beta1_istio_cni_gke.yaml) one in case you use GKE) to setup CNI with the operator.

Additionally, we are looking into [Pod Security Policies](https://kubernetes.io/docs/concepts/policy/pod-security-policy/) right now (as it was requested in #233) and we will soon provide at least a guide on how to use the operator with PSP enabled. Stay tuned!

allamand:
@Laci21 thanks a lot for opensourcing your work and for the pointers ;)

Laci21:
> @Laci21 thanks a lot for opensourcing your work and for the pointers ;)

That's what we do :)

No problem, good luck with them and let us know if you have any issues!

PR #258, opened 2019-07-09
Add Locality Load Balancing support

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

Adds support for locality based load balancing in Istio

### Checklist

- [x] Implementation tested


pkg/resources/common/configmap.go
```diff
@@ -117,6 +117,24 @@ func (r *Reconciler) meshConfig() string {
 	return string(marshaledConfig)
 }
 
+func (r *Reconciler) getLocalityLBConfiguration() *istiov1beta1.LocalityLBConfiguration {
+	var localityLbConfiguration *istiov1beta1.LocalityLBConfiguration
+
+	if !util.PointerToBool(r.Config.Spec.LocalityLB.Enabled) {
```
Laci21: Shouldn't we check first `if r.Config.Spec.LocalityLB != nil`? Because if it is `r.Config.Spec.LocalityLB.Enabled` would fail, isn't it?

pkg/resources/common/configmap.go
```diff
@@ -117,6 +117,24 @@ func (r *Reconciler) meshConfig() string {
...
+		return localityLbConfiguration
+	}
+
+	if r.Config.Spec.LocalityLB != nil {
+		localityLbConfiguration = r.Config.Spec.LocalityLB.DeepCopy()
+		localityLbConfiguration.Enabled = nil
```
Laci21: Why does it need to be set to `nil`?

waynz0r: Otherwise the `enabled: ` will be set to the meshconfig as well.

pkg/resources/common/configmap.go
waynz0r: Not sure actually, but i added a condition, just to be on the safe side.

Laci21: Thanks!

PR #296, opened 2019-09-14
Experimental Mixerless http telemetry support

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

- Experimental [Mixerless http telemetry](https://github.com/istio/istio/wiki/Mixerless-HTTP-Telemetry) support
### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/resources/mixerlesstelemetry/mixerlesstelemetry.go
+++ b/pkg/resources/mixerlesstelemetry/mixerlesstelemetry.go
@@ -1,0 +1,184 @@
+/*
+Copyright 2019 Banzai Cloud.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mixerlesstelemetry
+
+import (
+	"github.com/MakeNowJust/heredoc"
+	"github.com/go-logr/logr"
+	"github.com/goph/emperror"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	istiov1beta1 "github.com/banzaicloud/istio-operator/pkg/apis/istio/v1beta1"
+	"github.com/banzaicloud/istio-operator/pkg/k8sutil"
+	"github.com/banzaicloud/istio-operator/pkg/resources"
+	"github.com/banzaicloud/istio-operator/pkg/util"
+)
+
+const (
+	componentName = "mixerless-telemetry"
+)
+
+type Reconciler struct {
+	resources.Reconciler
+	dynamic dynamic.Interface
+}
+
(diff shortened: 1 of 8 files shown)


pkg/resources/mixerlesstelemetry/mixerlesstelemetry.go
```diff
@@ -0,0 +1,184 @@
...
+		},
+		Owner: r.Config,
+	}
+}
+
+func (r *Reconciler) statsEnvoyFilter() *k8sutil.DynamicObject {
+	filterInbound := map[string]interface{}{
+		"filterConfig": map[string]interface{}{
+			"configuration": heredoc.Doc(`
+{
```
pepov: I beleive this could be indented like it was a code block, heredoc should remove the obsolete whitespaces

pkg/resources/mixerlesstelemetry/mixerlesstelemetry.go
```diff
@@ -0,0 +1,184 @@
...
+			return emperror.WrapWith(err, "failed to reconcile dynamic resource", "resource", o.Gvr)
+		}
+	}
+
+	log.Info("Reconciled")
+
+	return nil
+}
+
+func (r *Reconciler) metaexchangeEnvoyFilter() *k8sutil.DynamicObject {
```
Laci21: Did you consider moving these envoyfilter functions to separate files?

pkg/resources/mixerlesstelemetry/mixerlesstelemetry.go
```diff
@@ -0,0 +1,184 @@
...
+			"index": "FIRST",
+		},
+		"listenerMatch": map[string]interface{}{
+			"listenerProtocol": "HTTP",
+			"listenerType":     "SIDECAR_INBOUND",
+		},
+	}
+
+	var filterOutbound map[string]interface{}
+	filterOutbound, _ = util.Map(filterInbound)
```
Laci21: There could be a helper function implemented with one string parameter to generate these filters.

Laci21:
If `mixerlessTelemetry.enabled` is set to to `true` in the Istio CR, shouldn't we make sure that mixer is not launched even if it would be defined in a user's Istio CR?

So I'm thinking checking it somewhere here: https://github.com/banzaicloud/istio-operator/blob/bec4157ee49137f077a760fcb3b8b64510acca12/pkg/resources/mixer/mixer.go#L70.

waynz0r:
> If `mixerlessTelemetry.enabled` is set to to `true` in the Istio CR, shouldn't we make sure that mixer is not launched even if it would be defined in a user's Istio CR?
> 

Since the new feature is experimental and also perfectly capable of running side-by-side with mixer telemetry I would not make them mutually exclusive. You can test the new feature out while still having mixer based telemetry to compare data for example.

pkg/resources/mixerlesstelemetry/mixerlesstelemetry.go
waynz0r: 👍

pkg/resources/mixerlesstelemetry/mixerlesstelemetry.go
waynz0r: Yeah, good idea!

Laci21:
> > If `mixerlessTelemetry.enabled` is set to to `true` in the Istio CR, shouldn't we make sure that mixer is not launched even if it would be defined in a user's Istio CR?
> 
> Since the new feature is experimental and also perfectly capable of running side-by-side with mixer telemetry I would not make them mutually exclusive. You can test the new feature out while still having mixer based telemetry to compare data for example.

Ok, if these features can coexist, then it's all good, thanks!

PR #300, opened 2019-09-25
Added support for setting domain (global.proxy.clusterDomain equivalent)

### LINKED ISSUE #101: [Istio 1.1] Configuration: domain

qvicksilver:
| Q               | A
| --------------- | ---
| Bug fix?             | no
| New feature?    |yes
| API breaks?       |no
| Deprecations?  |no
| Related tickets |fixes #101
| License             |Apache 2.0

### What's in this PR?
Enables setting the equivalent of global.proxy.clusterDomain from the Istio Helm Chart. Chose to default to "cluster.local" as in upstream. 

### Checklist

- [x] Implementation tested


waynz0r:
Thanks for the PR @qvicksilver!

At first glance it looks fine, but I think we should use the `clusterDomain` every place where we currently use fixed `cluster.local`. For example: https://github.com/banzaicloud/istio-operator/blob/13af2d5006adac1b3b09281cb1e2bf16c536e4bd/pkg/resources/pilot/meshexpansion.go#L79

qvicksilver:
> Thanks for the PR @qvicksilver!
> 
> At first glance it looks fine, but I think we should use the `clusterDomain` every place where we currently use fixed `cluster.local`. For example:
> 
> https://github.com/banzaicloud/istio-operator/blob/13af2d5006adac1b3b09281cb1e2bf16c536e4bd/pkg/resources/pilot/meshexpansion.go#L79

Sounds sane, I'll get on to it!

waynz0r:
@qvicksilver,

can you please rebase?

PR #301, opened 2019-09-25
Added support for setting trust domain (global.trustDomain equivalent)

### LINKED ISSUE #100: [Istio 1.1] Configuration: trust domain

qvicksilver:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #100

### What's in this PR?
Enables setting the equivalent of global.trustDomain from the Istio Helm Chart. Chose to default to "cluster.local" where that was done previously.

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/resources/citadel/deployment.go
+++ b/pkg/resources/citadel/deployment.go
@@ -41,6 +41,7 @@ func (r *Reconciler) deployment() runtime.Object {
 		fmt.Sprintf("--citadel-storage-namespace=%s", r.Config.Namespace),
 		fmt.Sprintf("--custom-dns-names=istio-pilot-service-account.%[1]s:istio-pilot.%[1]s", r.Config.Namespace),
 		"--monitoring-port=15014",
+		fmt.Sprintf("--trust-domain=%s", r.Config.Spec.TrustDomain),
 	)
 
 	if r.Config.Spec.Citadel.CASecretName == "" {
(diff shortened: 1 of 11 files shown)


pkg/resources/citadel/deployment.go
```diff
@@ -77,6 +77,13 @@ func (r *Reconciler) deployment() runtime.Object {
 		)
 	}
 
+	if r.Config.Spec.TrustDomain != "" {
+		args = append(args,
```
waynz0r: Could not we use this setting with a previously set default `cluster.local` without the condition?

qvicksilver: Probably, I just didn't want to go from implicit to explicit without a discussion about it

Laci21:
The PR looks great, there is just one place where I think it was missed earlier to add the `trustDomain` arg, namely in the Mixer deployment where it should be set according to this: https://github.com/istio/istio/blob/release-1.3/install/kubernetes/helm/istio/charts/mixer/templates/deployment.yaml#L136-L138

Could you please add it here as well: https://github.com/banzaicloud/istio-operator/blob/release-1.3/pkg/resources/mixer/deployment.go#L279 ?

Otherwise, I think the PR will be good to be merged.

PR #306, opened 2019-09-30
Istio 1.3.1 support

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

Add support for Istio 1.3.1

### DIFF
--- a/pkg/apis/istio/v1beta1/defaults.go
+++ b/pkg/apis/istio/v1beta1/defaults.go
@@ -140,6 +140,12 @@ func SetDefaults(config *Istio) {
 	if config.Spec.Pilot.EnableProtocolSniffing == nil {
 		config.Spec.Pilot.EnableProtocolSniffing = util.BoolPointer(true)
 	}
+	if config.Spec.Pilot.EnableProtocolSniffingOutbound == nil {
+		config.Spec.Pilot.EnableProtocolSniffingOutbound = config.Spec.Pilot.EnableProtocolSniffing
+	}
+	if config.Spec.Pilot.EnableProtocolSniffingInbound == nil {
+		config.Spec.Pilot.EnableProtocolSniffingInbound = config.Spec.Pilot.EnableProtocolSniffing
+	}
 	// Citadel config
 	if config.Spec.Citadel.Enabled == nil {
 		config.Spec.Citadel.Enabled = util.BoolPointer(true)
--- a/pkg/apis/istio/v1beta1/defaults.go
+++ b/pkg/apis/istio/v1beta1/defaults.go
@@ -144,7 +144,7 @@ func SetDefaults(config *Istio) {
 		config.Spec.Pilot.EnableProtocolSniffingOutbound = config.Spec.Pilot.EnableProtocolSniffing
 	}
 	if config.Spec.Pilot.EnableProtocolSniffingInbound == nil {
-		config.Spec.Pilot.EnableProtocolSniffingInbound = config.Spec.Pilot.EnableProtocolSniffing
+		config.Spec.Pilot.EnableProtocolSniffingInbound = util.BoolPointer(false)
 	}
 	// Citadel config
 	if config.Spec.Citadel.Enabled == nil {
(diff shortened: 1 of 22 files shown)


pkg/resources/nodeagent/daemonset.go
```diff
@@ -75,6 +75,10 @@ func (r *Reconciler) daemonSet() runtime.Object {
 									Name:  "Trust_Domain",
```
Laci21: Why are you keeping this one?

pkg/apis/istio/v1beta1/defaults.go
```diff
@@ -140,6 +140,12 @@ func SetDefaults(config *Istio) {
 	if config.Spec.Pilot.EnableProtocolSniffing == nil {
 		config.Spec.Pilot.EnableProtocolSniffing = util.BoolPointer(true)
 	}
+	if config.Spec.Pilot.EnableProtocolSniffingOutbound == nil {
+		config.Spec.Pilot.EnableProtocolSniffingOutbound = config.Spec.Pilot.EnableProtocolSniffing
+	}
+	if config.Spec.Pilot.EnableProtocolSniffingInbound == nil {
+		config.Spec.Pilot.EnableProtocolSniffingInbound = config.Spec.Pilot.EnableProtocolSniffing
```
Laci21: This seems to be `false` by default: https://github.com/istio/istio/blob/6d0953f791f76f614c2478f463eef166ea399105/pilot/pkg/features/pilot.go#L183

Also, we might consider adding the description from the link to the variables like: `If enabled, protocol sniffing will be used for inbound listeners whose port protocol is not specified or unsupported`

pkg/resources/pilot/deployment.go
```diff
@@ -157,6 +157,20 @@ func (r *Reconciler) containers() []apiv1.Container {
 		})
 	}
 
+	if util.PointerToBool(r.Config.Spec.Pilot.EnableProtocolSniffingOutbound) {
+		discoveryContainer.Env = append(discoveryContainer.Env, apiv1.EnvVar{
+			Name:  "PILOT_ENABLE_PROTOCOL_SNIFFING_FOR_OUTBOUND",
+			Value: "true",
```
Laci21: Are u sure we always want to set this to `true`? If `util.PointerToBool(r.Config.Spec.Pilot.EnableProtocolSniffingOutbound)` is `false` then we don't set this env var, but won't it be initialized to be `true` anyway: https://github.com/istio/istio/blob/6d0953f791f76f614c2478f463eef166ea399105/pilot/pkg/features/pilot.go#L177?

pkg/resources/nodeagent/daemonset.go
waynz0r: I left it there so the operator to be backward compatible.

pkg/apis/istio/v1beta1/defaults.go
waynz0r: 👍

pkg/resources/pilot/deployment.go
waynz0r: 👍

pkg/resources/pilot/deployment.go
```diff
@@ -150,12 +150,21 @@ func (r *Reconciler) containers() []apiv1.Container {
...
-	if util.PointerToBool(r.Config.Spec.Pilot.EnableProtocolSniffing) {
-		discoveryContainer.Env = append(discoveryContainer.Env, apiv1.EnvVar{
-			Name:  "PILOT_ENABLE_PROTOCOL_SNIFFING",
-			Value: "true",
-		})
-	}
+	// TODO - remove it in the next minor -- left it here for backward compatibility
+	discoveryContainer.Env = append(discoveryContainer.Env, apiv1.EnvVar{
+		Name:  "PILOT_ENABLE_PROTOCOL_SNIFFING",
+		Value: strconv.FormatBool(util.PointerToBool(r.Config.Spec.Pilot.EnableProtocolSniffingOutbound)),
```
Laci21:
```suggestion
		Value: strconv.FormatBool(util.PointerToBool(r.Config.Spec.Pilot.EnableProtocolSniffing)),
```

PR #307, opened 2019-09-30
galley validation webhook fix

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

Fixes galley validation on/off by adding webhook removal directly to the operator

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/resources/galley/configmap.go
+++ b/pkg/resources/galley/configmap.go
@@ -38,20 +38,20 @@ func (r *Reconciler) configMap() runtime.Object {
 	}
 
 	if util.PointerToBool(r.Config.Spec.Galley.ConfigValidation) {
-		configmap.Data["validatingwebhookconfiguration.yaml"] = r.validatingWebhookConfig(r.Config.Namespace)
+		configmap.Data["validatingwebhookconfiguration.yaml"] = r.validatingWebhookConfig()
 	}
 
 	return configmap
 }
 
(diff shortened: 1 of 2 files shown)


pkg/resources/galley/configmap.go
```diff
@@ -38,28 +38,28 @@ func (r *Reconciler) configMap() runtime.Object {
...
-func (r *Reconciler) validatingWebhookConfig(ns string) string {
+func (r *Reconciler) validatingWebhook() runtime.Object {
 	fail := admissionv1beta1.Fail
 	se := admissionv1beta1.SideEffectClassNone
-	webhook := admissionv1beta1.ValidatingWebhookConfiguration{
+
+	return &admissionv1beta1.ValidatingWebhookConfiguration{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      webhookName,
-			Namespace: ns,
```
Laci21: I guess the namespace is omitted from here deliberately, but just to double check, is it? :)

waynz0r: Yes, the admission webhooks are cluster scoped resources.

Laci21: Thanks!

PR #347, opened 2019-12-08
Multiple mesh gateways support

### LINKED ISSUE #271: Support multiple ingress gateways
**Problem statement:**
The current version of the operator has the Istio ingress gateway configuration in the main `Istio` custom resource. It means that only one ingress gateway can be configured and reconciled through the operator, but it is a common use-case to have multiple ingress gateways in an environment.

**Possible workaround:**
Configure the Istio control plane (and one ingress gateway) through the operator and manage the rest of the gateways manually by installing them with Helm, or a simple `kubectl apply`.

**Solution:**
Decouple the ingress gateway configuration from the rest of the control plane components by creating a new custom resource definition (`IstioIngressGateway`) that holds these properties, and a new controller in the operator that manages this component. Adding a new CR would create a new gateway, deleting the CR would delete that gateway.
The current APIs and controllers are generated by [Kubebuilder v1](https://book-v1.book.kubebuilder.io), the new CRDs and controllers should use the same version.

Make sure that the current Istio control plane CRD remains backward compatible, so an ingress gateway could still be created through the main CR.

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #271 

### What's in this PR?

This PR adds a new CRD (`MeshGateway`) and controller for it to the operator. With the `MeshGateway` resources it is possible to define as many Istio ingress/egress gateways as needed.

Also from now on the main ingress/egress gateway defined in the `Istio` CR is also not created directly, but by using the `MeshGateway` resource.

### Why?

The current version of the operator has the Istio ingress/egress gateway configuration in the main Istio custom resource. It means that only one ingress gateway can be configured and reconciled through the operator, but it is a common use-case to have multiple ingress gateways in an environment. Having a separate CRD to define a mesh gateway solves that problem and it also provides a way through RBAC rules to control who can create those gateways.

### Additional context

Example custom ingress gateway:

```
apiVersion: istio.banzaicloud.io/v1beta1
kind: MeshGateway
metadata:
  name: example-mesh-ingress
  namespace: default
spec:
  labels:
    example: label
  type: ingress
  serviceType: LoadBalancer
  ports:
  - name: status-port
    port: 15020
    protocol: TCP
  - name: http2
    port: 80
    protocol: TCP
```

```
$ kubectl -n default get mgw
NAME                   TYPE      SERVICE TYPE   STATUS      ERROR   AGE
example-mesh-ingress   ingress   LoadBalancer   Available           4m
```

### Checklist

- [x] Implementation tested in single cluster
- [x] Implementation tested in single mesh multi cluster


pkg/resources/gateways/deployment.go
```diff
@@ -342,16 +320,17 @@ func (r *Reconciler) envVars(gw string, gwConfig *istiov1beta1.GatewayConfigurat
 		})
 	}
 
-	if util.PointerToBool(r.Config.Spec.MeshExpansion) && gw == ingress && r.Config.Spec.GetNetworkName() != "" {
+	// TODO ingress
```
Laci21: left TODO?

pkg/resources/gateways/deployment.go
```diff
@@ -424,21 +403,21 @@ func (r *Reconciler) envVars(gw string, gwConfig *istiov1beta1.GatewayConfigurat
...
 	vms := []apiv1.VolumeMount{
 		{
 			Name:      "istio-certs",
 			MountPath: "/etc/certs",
 			ReadOnly:  true,
 		},
 		{
-			Name:      fmt.Sprintf("%s-certs", gw),
-			MountPath: fmt.Sprintf("/etc/istio/%s-certs", gw),
+			Name:      fmt.Sprintf("%s-certs", "ingressgateway"),
```
Laci21: You could consider using a const var for `"ingressgateway"` in this file.

pkg/apis/istio/v1beta1/meshgateway_defaults.go
```diff
@@ -0,0 +1,47 @@
...
+		gw.Spec.ReplicaCount = util.IntPointer(defaultReplicaCount)
+	}
+	if gw.Spec.MinReplicas == nil {
+		gw.Spec.MinReplicas = util.IntPointer(defaultReplicaCount)
+	}
+	if gw.Spec.MaxReplicas == nil {
+		gw.Spec.MaxReplicas = util.IntPointer(defaultReplicaCount)
+	}
+	if gw.Spec.Resources == nil {
+		gw.Spec.Resources = defaultProxyResources
```
Laci21: `defaultProxyResources` is used deliberately here?

PR #359, opened 2020-01-30
add TLS support for Zipkin tracing endpoint

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

Add support to specify TLS settings for Zipkin tracing endpoint.

### Why?

Needs TLS settings for the configured Zipkin endpoint to support multi-cluster tracing.

### Additional context

This feature also needs support for this feature on the proxy side. That is currently means specific image from Banzai Cloud.

### Checklist

- [x] Implementation tested


pkg/resources/sidecarinjector/configmap.go
```diff
@@ -48,6 +49,11 @@ func (r *Reconciler) getValues() string {
 		}...)
 	}
 
+	var zipkinTLSSettingsJSON []byte
+	if r.Config.Spec.Tracing.Tracer == v1beta1.TracerTypeZipkin {
```
Laci21: Shouldn't we check if `r.Config.Spec.Tracing.Zipkin.TLSSettings` is nil?

PR #370, opened 2020-02-12
Generate Istio CRDs from YAML manifests

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

Replaces go code with yaml manifests based CRD resource generation.

### Why?

The previous implementation was missing the lately added CRD validations, command aliases and was hard to maintain.

### Checklist

- [x] Implementation tested


pkg/crds/crds.go
```diff
@@ -26,142 +28,120 @@ import (
...
+		var crd *extensionsobj.CustomResourceDefinition
+		var ok bool
+		if crd, ok = obj.(*extensionsobj.CustomResourceDefinition); !ok {
+			continue
+		}
+
+		crd.Status = extensionsobj.CustomResourceDefinitionStatus{}
+		crd.SetGroupVersionKind(schema.GroupVersionKind{})
+		labels := crd.GetLabels()
+		labels[createdByLabel] = "istio-operator"
```
Laci21: Could we move "istio-operator" to a const var?

pkg/crds/crds.go
```diff
@@ -26,142 +28,120 @@ import (
...
-		crd("ServiceRole", "ServiceRoles", crdConfigs[Rbac], "mixer", "istio.io.mixer", "rbac", extensionsobj.NamespaceScoped),
-		crd("ServiceRoleBinding", "ServiceRoleBindings", crdConfigs[Rbac], "mixer", "istio.io.mixer", "rbac", extensionsobj.NamespaceScoped),
-		crd("adapter", "adapters", crdConfigs[Policy], "mixer", "istio.io.mixer", "rbac", extensionsobj.NamespaceScoped),
-		crd("instance", "instances", crdConfigs[Policy], "mixer", "instance", "mixer-instance", extensionsobj.NamespaceScoped),
-		crd("template", "templates", crdConfigs[Policy], "mixer", "template", "mixer-template", extensionsobj.NamespaceScoped),
-		crd("handler", "handlers", crdConfigs[Policy], "mixer", "handler", "mixer-handler", extensionsobj.NamespaceScoped),
-		crd("Sidecar", "Sidecars", crdConfigs[Networking], "istio-pilot", "", "", extensionsobj.NamespaceScoped),
-		crd("AuthorizationPolicy", "authorizationpolicies", crdConfigs[Security], "isito-pilot", "", "security", extensionsobj.NamespaceScoped),
+func (r *CRDReconciler) LoadCRDs() error {
+	for _, file := range []string{"crd-10.yaml", "crd-11.yaml", "crd-14.yaml"} {
```
Laci21: Alternatively file names in the repo could be listed here, instead of hard-coding them. Later if new files are added they would not be forgotten in that case.

pkg/crds/crds.go
waynz0r: 👍

pkg/crds/crds.go
waynz0r: 👍

pkg/crds/crds.go
```diff
@@ -60,15 +61,24 @@ func New(cfg *rest.Config, crds ...*extensionsobj.CustomResourceDefinition) (*CR
 }
 
 func (r *CRDReconciler) LoadCRDs() error {
-	for _, file := range []string{"crd-10.yaml", "crd-11.yaml", "crd-14.yaml"} {
-		f, err := istio_crds.CRDs.Open(file)
+	dir, err := istio_crds.CRDs.Open("/")
+	if err != nil {
+		return err
+	} else {
```
Laci21: I guess lint will cry about this else block.

pkg/crds/crds.go
Laci21: Great, thanks!

Issue #294, opened 2019-09-11
Supporting custom env vars for components


mousavian:
**Is your feature request related to a problem? Please describe.**
Some components of istio come with some tuning flags that can be set through ENV variables. 
(eg [pilot](https://istio.io/docs/reference/commands/pilot-discovery/#envvars), [galley](https://istio.io/docs/reference/commands/galley/#envvars), etc.) as well as some env vars that affect [Golang runtime](https://dave.cheney.net/2015/11/29/a-whirlwind-tour-of-gos-runtime-environment-variables).
In my case, I need to provide `GOMAXPROCS` for all components, specifically pilot/discovery container as we see a lots of CPU throttling happening there.

Similar to #216  but in more generic way.

**Describe the solution you'd like to see**
I would like to be able to set env variables as key value in my kustomization for each component under `Kind: Istio`. Similar to what [istio/installer using Helm](https://github.com/istio/istio/blob/e751268/install/kubernetes/helm/istio/charts/pilot/templates/deployment.yaml#L104-L109) is providing.

**Describe alternatives you've considered**

**Additional context**
Manually adding `GOMAXPROCS` to the right value in pilot deployment helped us to reduce cpu throttling and memory consumption at the same time.

<img width="558" alt="Screen Shot 2019-09-11 at 13 07 10" src="https://user-images.githubusercontent.com/317322/64665226-2061b280-d495-11e9-9b31-c036a1188a2e.png">

```
rate(container_cpu_cfs_throttled_seconds_total{namespace="$namespace",container_name="$container"}[5m])
```

Laci21:
Hi @mousavian,

This feature would make sense for sure.

We will do our best to implement this feature in the next 2-3 weeks, but in around two weeks we will have a [Backyards](https://banzaicloud.com/blog/istio-the-easy-way/) release, until then we have limited capacity, so we cannot promise anything. If you need the feature sooner and you are up for it, we are more than happy to accept PRs as you already know. Otherwise, we will try to resolve this as soon as we can.

BoOgatti:
Hi,
In my case it will be useful for enabling HTTP/1.0 (PILOT_HTTP10: 1). Now I have to doing it manually

Laci21:
Hi @mousavian and @BoOgatti,

Sorry for the unusually long delay with this feature, but finally it was implemented in #374, courtesy of @waynz0r!

You can use the new `additionalEnvVars` field under each component to set any custom environment variables for the Istio control plane components.

[...]

PR #380, opened 2020-02-25
Support additional env variables in MeshGateway resource

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| API breaks?     |  no|yes
| Deprecations?   | no|yes
| Related tickets | #294  

### What's in this PR?

Add support for additional env variables to `MeshGateway` resource.

### Why?

Currently is not possible to set additional environment variables, but it is a necessity for advanced configuration.

### Additional context

Variables are merged by name,  new variables are simply appended.
This is an advanced feature which can lead to misconfiguration if not used properly.

### Checklist

- [x] Implementation tested

### DIFF
--- a/config/samples/istio_v1beta1_meshgateway.yaml
+++ b/config/samples/istio_v1beta1_meshgateway.yaml
@@ -25,7 +25,13 @@ spec:
     - port: 15443
       targetPort: 15443
       name: tls
-  additionalEnvVars: []
+#   additionalEnvVars:
+#   - name: ADDITIONAL_VARIABLE
+#     value: "value"
+#   - name: META_POD_NAME
+#     valueFrom:
+#       fieldRef:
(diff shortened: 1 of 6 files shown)


config/samples/istio_v1beta1_meshgateway.yaml
```diff
@@ -25,6 +25,7 @@ spec:
     - port: 15443
       targetPort: 15443
       name: tls
+  additionalEnvVars: []
```
Laci21: We could add a commented out version of an example of env vars here (and to the istio CR as well) so that users would know what is the structure they need to provide for env vars.

waynz0r: 👍

PR #381, opened 2020-02-26
Support Istio 1.5.0

Laci21:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

- Support for Istio `1.5.x` and make `1.5.0` the default version on the `release-1.5` branch
- Changes: https://github.com/istio/istio/compare/1.4.5...1.5.0

### To Do

- [x] Test: Single cluster
- [x] Test: Upgrade
- [x] Istiod
- [x] Mixerless telemetry
- [x] Add sample CR to multiple component setup 
- [x] SDS
- [x] Bump to 1.5.0 tag when released (check if there are any relevant changes compared to beta tag)


pkg/resources/base/common.go
```diff
@@ -27,11 +27,16 @@ import (
 )
 
 const (
-	componentName      = "common"
-	istioReaderName    = "istio-reader"
-	IstioConfigMapName = "istio"
+	componentName                 = "common"
+	istioReaderName               = "istio-reader"
+	istioReaderServiceAccountName = istioReaderName + "service-account"
```
khernyo: `istioReaderServiceAccountName = istio-readerservice-account`
Shouldn't it have a `-` in between?

Laci21: True dat, nice catch, fixed, thank you! :)

PR #401, opened 2020-03-18
configurable wasm runtime for proxies

martonsereg:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
If wasm is enabled, EnvoyFilters need to be configured differently, to pick the wasm extensions from the filesystem.

### DIFF
--- a/pkg/resources/mixerlesstelemetry/meta_exchange_filter.go
+++ b/pkg/resources/mixerlesstelemetry/meta_exchange_filter.go
@@ -17,13 +17,17 @@ limitations under the License.
 package mixerlesstelemetry
 
 import (
+	"fmt"
+
+	"github.com/banzaicloud/istio-operator/pkg/util"
 	"github.com/ghodss/yaml"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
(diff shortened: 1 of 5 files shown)


pkg/resources/mixerlesstelemetry/meta_exchange_filter.go
```diff
@@ -96,10 +98,26 @@ const TCPMetadataExchangeFilterYAML = `
...
                 protocol: istio-peer-exchange
 `
+	metaExchangeWasmLocal   = "filename: /etc/istio/extensions/metadata-exchange-filter.wasm"
+	metaExchangeNoWasmLocal = "inline_string: envoy.wasm.metadata_exchange"
+)
 
 func (r *Reconciler) metaexchangeEnvoyFilter() *k8sutil.DynamicObject {
+
+	wasmEnabled := false
+	if r.Config.Spec.ProxyWasm.Enabled != nil {
```
Laci21: We have a `util.PointerToBool` function for this.

PR #418, opened 2020-04-23
circleci: build docker images

asdwsda:
- [ ] Disable autobuild on dockerhub before merge

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,32 @@
 version: 2.1
 orbs:
   helm: banzaicloud/helm@0.0.3
+  docker: banzaicloud/docker@0.0.3
+
+commands:
+  publish-with-latests:
+    steps:
+      - docker/tag:
+          target-image: banzaicloud/istio-operator
+          target-tag: ${CIRCLE_TAG}
+      - docker/push:
+          image: banzaicloud/istio-operator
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -94,6 +120,33 @@ workflows:
...
+          name: Publish master docker image
+          executor: docker/machine
+          context: dockerhub
+          image: banzaicloud/istio-operator
+          tag: master
+          requires:
+            - build
+          filters:
+            branches:
+              only: master
```
Laci21, line 133: We don't use `master`, I think this is not needed.

Laci21:
> Are the specific tags pushed from somewhere?

Ah, I guess they're pushed from here: https://github.com/banzaicloud/istio-operator/pull/418/files#diff-1d37e48f9ceff6d8030570cd36286a61R12 nvm

.circleci/config.yml
khernyo: I think this job could be run for all release-* branches instead.
I'm not sure what would be a good docker tag, though. E.g. release-1.5 is not a good tag. Maybe dev-1.5?

Laci21: I don't think that's needed, imho there is no need to push new dev docker images on every single release branch push, it's just enough on tags.

PR #459, opened 2020-06-28
Migrate to go modules

### LINKED ISSUE #456: Errors you may encounter when upgrading the library
### github.com/mholt/caddy  
-Now Moved To: github.com/caddyserver/caddy
-Latest Version: v2.1.0   (Latest commit c9049bd yesterday)
-Where did you use it: 
https://github.com/banzaicloud/istio-operator/search?q=mholt%2Fcaddy&unscoped_q=mholt%2Fcaddy 
-Detail: 
>[github.com/caddyserver/caddy/go.mod](https://github.com/caddyserver/caddy/blob/v2.1.0/go.mod#L1)
```
module github.com/caddyserver/caddy/v2
go 1.14
require (
	github.com/Masterminds/sprig/v3 v3.1.0
	github.com/go-acme/lego/v3 v3.7.0
)  
```
>[github.com/caddyserver/caddy/caddyconfig/caddyfile/adapter.go](https://github.com/caddyserver/caddy/blob/v2.1.0/caddyconfig/caddyfile/adapter.go#L21)
```
package caddyfile
import (
	"github.com/caddyserver/caddy/v2"
	"github.com/caddyserver/caddy/v2/caddyconfig"
) 
```
This problem was introduced since **_caddyserver/caddy v1.0.4(committed aadda6e on 16 Nov 2019)_** . If you try to upgrade **caddyserver/caddy** to version **v1.0.4** and **above**, you will get an error--- **_no package exists at "github.com/caddyserver/caddy/v2"_**

**NOTICE HERE**  The `github.com/mholt/caddy` makes it very difficult for downstream module users to use `banzaicloud/istio-operator`. Force downstream module users to use vendor if they want to use `banzaicloud/istio-operator`.

## I investigated the libraries  (**caddyserver/caddy >= v1.0.4**)  release information and found the root cause of this issue is that---- 
1.	These dependencies all added [Go modules](https://github.com/golang/go/wiki/Modules) in the recent versions.

2.	They all comply with the specification of "Releasing Modules for v2 or higher" available in the Modules documentation. Quoting the specification:
> A package that has migrated to Go Modules _must_ include the major version in the import path to reference any v2+ modules. For example,  Repo github.com/my/module migrated to Modules on version v3.x.y.  Then this repo should declare its module path with MAJOR version suffix "/v3"  (e.g., module `github.com/my/module/v3`), and its downstream project should use `"github.com/my/module/v3/mypkg"` to import this repo’s package.

3.	This "github.com/my/module/v3/mypkg" is not the `physical path`. So earlier versions of Go (including those that don't have minimal module awareness) plus all tooling (like dep, glide, govendor, etc) don't have `minimal module awareness` as of now and therefore don't handle import paths correctly See [golang/dep#1962](https://github.com/golang/dep/issues/1962), [golang/dep#2139](https://github.com/golang/dep/issues/2139).
> Note: creating a new branch is not required. If instead you have been previously releasing on master and would prefer to tag v3.0.0 on master, that is a viable option. (However, be aware that introducing an incompatible API change in master can cause issues for non-modules users who issue a go get -u given the go tool is not aware of semver prior to Go 1.11 or when module mode is not enabled in Go 1.11+).
> Pre-existing dependency management solutions such as dep currently can have problems consuming a v2+ module created in this way. See for example dep#1962.
https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher

### 1. Migrate to Go Modules.
Go Modules is the general trend of ecosystem, If you want a better upgrade package experience, migrating to Go Modules is a good choice.

### 2. Keep using the dependency manage tool (like dep, glide, govendor, etc).

Because the import paths have different meanings between the projects adopting module mechanism and the non-module mechanism. 

* https://github.com/golang/go/wiki/Modules#semantic-import-versioning
* https://golang.org/cmd/go/#hdr-Module_compatibility_and_semantic_versioning
* https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher

If you prefer to use the dependencies manage tool (like dep, glide, govendor, etc), a textbook example provided by repo `github.com/moby/moby` can prevent this type of issues.
https://github.com/moby/moby/blob/master/VENDORING.md
https://github.com/moby/moby/blob/master/vendor.conf
In the vendor directory, `github.com/moby/moby` adds the **/vN** subdirectory in the corresponding dependencies.
**This will help more downstream modules to work well with your package, otherwise they will be stuck in github.com/caddyserver/caddy v1.0.5.**

**By the way**, have you ever encountered such errors when upgading the libraries? Do you plan to upgrade the libraries in near future?
Hope this issue report can help you ^_^
Thank you very much for your attention.

Best regards,
Kate

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #456

### What's in this PR?

Migrate from `dep` to `go modules`.

### Checklist

- [x] Implementation tested


sagikazarmark:
🎉

PR #505, opened 2020-09-02
Restructure CRDs in Helm chart

Laci21:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

1. Move CRDs from `templates` to `crds` dir
1. Delete CRDs for earlier versions
1. Remove templating from CRD yamls
1. Rename CRD files

### Why?

1. Adjust to Helm3 chart structure: this way it is possible to install charts without the CRDs and more importantly during uninstall the CRDs are not deleted. Source: [https://helm.sh/docs/topics/charts/#custom-resource-definitions-crds](https://helm.sh/docs/topics/charts/#custom-resource-definitions-crds)
1. Our CRDs should be backward compatible, so the newest one can be used for earlier versions as well
1. It is forbidden to use templating for CRDs in Helm3
1. No need to have version names for CRDs anymore

### Additional context

This way two operator versions can be installed next to each other and then when removing one of them the CRDs are not removed.

### Checklist

- [x] Implementation tested


turip (approved):

waynz0r (approved):
thanks @Laci21!

PR #507, opened 2020-09-04
Fix create only ingress/egress reconcile

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

Removes owner reference from the generated main ingress/egress MGWs if the `createOnly` property is true.

### Why?

Having the owner references causes removal of the child resource, that is unwanted when the `createOnly` is true for the generated gateways.

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/k8sutil/resource.go
+++ b/pkg/k8sutil/resource.go
@@ -52,6 +52,14 @@ func Reconcile(log logr.Logger, client runtimeClient.Client, desired runtime.Obj
 	}
 	if apierrors.IsNotFound(err) {
 		if desiredState != DesiredStateAbsent {
+			should, err := desiredState.ShouldCreate(current)
+			if err != nil {
+				return emperror.WrapWith(err, "could not execute ShouldCreate func")
+			}
+			if !should {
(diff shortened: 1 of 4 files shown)


turip (approved):
🎆

pkg/k8sutil/resource.go
```diff
@@ -105,12 +121,19 @@ func Reconcile(log logr.Logger, client runtimeClient.Client, desired runtime.Obj
 
 			if err := client.Update(context.TODO(), desired); err != nil {
 				if apierrors.IsConflict(err) || apierrors.IsInvalid(err) {
+					should, err := desiredState.ShouldRecreate(current, desiredCopy)
+					if err != nil {
+						return emperror.WrapWith(err, "could not execute ShoudCreate func")
```
Laci21, line 126:
```suggestion
						return emperror.WrapWith(err, "could not execute ShoudReCreate func")
```

Laci21 (approved):

PR #521, opened 2020-09-24
Label namespace only if istio.io/rev label is not set

Laci21:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### Why?

Otherwise when upgrading to 1.7 Istio CP, the 1.6 operator might reconcile back the `istio-injection` label, which is not intended.

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/k8sutil/resource.go
+++ b/pkg/k8sutil/resource.go
@@ -185,7 +185,7 @@ func IsObjectChanged(oldObj, newObj runtime.Object, ignoreStatusChange bool) (bo
 }
 
 // ReconcileNamespaceLabelsIgnoreNotFound patches namespaces by adding/removing labels, returns without error if namespace is not found
-func ReconcileNamespaceLabelsIgnoreNotFound(log logr.Logger, client runtimeClient.Client, namespace string, labels map[string]string, labelsToRemove []string) error {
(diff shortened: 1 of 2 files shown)


pkg/k8sutil/resource.go
```diff
@@ -197,6 +197,13 @@ func ReconcileNamespaceLabelsIgnoreNotFound(log logr.Logger, client runtimeClien
 		return emperror.WrapWith(err, "getting namespace failed", "namespace", namespace)
 	}
 
+	if ns.Labels != nil {
```
khernyo, line 200: I don't think this is necessary

pkg/k8sutil/resource.go
```diff
@@ -197,6 +197,13 @@ func ReconcileNamespaceLabelsIgnoreNotFound(log logr.Logger, client runtimeClien
...
+		if _, ok := ns.Labels["istio.io/rev"]; ok {
+			log.V(1).Info("namespace has 'istio.io/rev' label, ignoring", "namespace", namespace)
```
khernyo, line 202:
```suggestion
			log.V(1).Info("namespace has 'istio.io/rev' label, ignoring namespace", "namespace", namespace)
```

khernyo (approved):

turip (approved):

waynz0r (approved):

PR #531, opened 2020-09-30
Ensure correct codegen versions

khernyo:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets |

### What's in this PR?
Same as #528, but for release-1.6 branch: `download-deps.sh` now saves the binaries with a version suffix and creates a symlink to the currently used version

### Why?
For reproducible builds

### Checklist
- [x] Implementation tested


Laci21 (approved):
lgtm, thanks @khernyo!

scripts/download-deps.sh
```diff
@@ -1,18 +1,38 @@
...
+
+function ensure-binary-version() {
+    local bin_name=$1
+    local bin_version=$2
+    local download_location=$3
+
+    local target_name=${bin_name}-${bin_version}
+    local link_path=${binpath}/${bin_name}
+
+    [ -e "${link_path}" ] && rm "${link_path}"
```
turip, line 19: Wouldn't it cause issues with the set -e if the link path does not exist?

khernyo: Nope, it's part of a compound command. I double checked by deleting one of the symlinks under bin. It's recreated properly.

turip (approved):

PR #535, opened 2020-10-12
Fix create only ingress/egress reconcile

Laci21:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

Cherry-pick of #507.

### What's in this PR?

Removes owner reference from the generated main ingress/egress MGWs if the `createOnly` property is true.

### Why?

Having the owner references causes removal of the child resource, that is unwanted when the `createOnly` is true for the generated gateways.

### Checklist

- [x] Implementation tested


khernyo (approved):

waynz0r (approved):
thanks @Laci21!

PR #553, opened 2020-12-07
Periodically reconcile mesh gateways with hostnames 

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

Add periodic reconcile for mesh gateways with hostnames.

### Why?

The IP addresses could change, the periodic reconcile will keep them up to date.

### DIFF
--- a/pkg/resources/gateways/externalip.go
+++ b/pkg/resources/gateways/externalip.go
@@ -35,7 +35,7 @@ func (r *Reconciler) GetGatewayAddress() ([]string, bool, error) {
 		Namespace: r.gw.Namespace,
 	}, &service)
 	if err != nil {
-		return nil, false, err
+		return nil, hasHostname, err
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/resources/gateways/externalip.go
```diff
@@ -25,22 +25,23 @@ import (
...
 	var ips []string
+	var hasHostname bool
 
 	err := r.Get(context.Background(), types.NamespacedName{
 		Name:      r.gatewayName(),
 		Namespace: r.gw.Namespace,
 	}, &service)
 	if err != nil {
-		return nil, err
+		return nil, false, err
```
khernyo, line 38: nit: this could be `hasHostname`, just to keep the pattern.
```suggestion
		return nil, hasHostname, err
```

khernyo (approved):

Laci21 (approved):

Issue #550, opened 2020-11-30
Use CA Certificate from Vault 


bonifaido:
**Is your feature request related to a problem? Please describe.**
It would be nice if the CA secret could be read from Vault instead of letting Istio generate one.

**Describe the solution you'd like to see**
An attribute block in the Istio CR would control the source of CA certificate.

`vault-env` could be used to authenticate and read the defined Secret from Vault and write the files to an ephemeral volume from where Istio can read them.

For Vault, the Secret path should be configurable which is `secret/data/pki/root` in the following example `initContainer` example, which should be the part of the solution we have discussed. Also, the Vault address and role should be possible to configure.

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: istiod
  labels:
    name: istiod
spec:
  initContainers:
  - name: vault-env
    image: ghcr.io/banzaicloud/vault-env:1.8.0
    env:
      - name: VAULT_ADDR
        value: https://vault.vault:8200
      - name: VAULT_ROLE
        value: istiod
      - name: ISTIO_CA_CERT
        value: vault:secret/data/pki/ca#certificate
      - name: ISTIO_CA_KEY
        value: vault:secret/data/pki/ca#private_key
      - name: VAULT_CACERT
        value: /vault/tls/ca.crt
    command:
      - /vault/vault-env
      - sh
      - -c
      - "echo $ISTIO_CA_CERT > ca-cert.pem; echo $ISTIO_CA_KEY > ca-key.pem"
    workingDir: /cacerts
    volumeMounts:
      - mountPath: /cacerts
        name: ca-cert
      - mountPath: /vault/tls/
        name: vault-tls
  volumes:
    - name: ca-cert
      emptyDir: {}
    - name: vault-tls
      projected:
        defaultMode: 420
        sources:
        - secret:
            name: vault-tls
            items:
            - key: ca.crt
              path: ca.crt
```

The `serviceAccount` of the `istiod` Pod has to be configured in Vault to be able to read the aforementioned path.

**Describe alternatives you've considered**
N/A

**Additional context**

PR #559, opened 2021-01-04
Support Istiod CA certificate injection from Vault

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | implements #550 

### What's in this PR?

Implements support to read/inject CA cert/key from Vault to Istiod.

### Why?

CA cert/key can be specified through a secret defined in the cluster, but it would be more secure to read it from Vault and inject it into the Istiod pod at runtime.

### Checklist

- [x] Implementation tested


baluchicken (approved):
thx @waynz0r

bonifaido (approved):
LGTM!

pkg/resources/istiod/deployment.go
```diff
@@ -205,6 +215,63 @@ func (r *Reconciler) containers() []apiv1.Container {
...
+	}
+
+	certChainPath := util.PointerToString(r.Config.Spec.Istiod.CA.Vault.CertPath)
+	if r.Config.Spec.Istiod.CA.Vault.CertChainPath != nil {
+		certChainPath = util.PointerToString(r.Config.Spec.Istiod.CA.Vault.CertChainPath)
+	}
+	vaultEnvContainer.Env = append(vaultEnvContainer.Env, apiv1.EnvVar{
+		Name:  "ISTIO_CA_CERT_CHAIN",
+		Value: certChainPath,
+	})
```
Laci21, line 270: Nit: `certChainPath` could be assigned to its appropriate value before initializing the `vaultEnvContainer` variable, that way the `ISTIO_CA_CERT_CHAIN` env var could be added to `vaultEnvContainer.Env` during initialization instead of appending to it later.

pkg/resources/istiod/deployment.go
```diff
@@ -290,7 +357,39 @@ func (r *Reconciler) volumes() []apiv1.Volume {
...
+		if util.PointerToBool(r.Config.Spec.Istiod.CA.Vault.Enabled) {
+			volumes = append(volumes, []apiv1.Volume{
+				{
+					Name: "cacerts",
+					VolumeSource: apiv1.VolumeSource{
+						EmptyDir: &apiv1.EmptyDirVolumeSource{},
+					},
+				},
+				{
+					Name: "vault-tls",
```
Laci21, line 369: Nit: String variables could be used for values, which are used multiple places e.g. volume names, env var names.

Laci21 (approved):
👍🏼

pkg/resources/istiod/deployment.go
waynz0r: 👍

pkg/resources/istiod/deployment.go
waynz0r: 👍

Laci21 (approved):

PR #566, opened 2021-01-14
Multi mesh improvements

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

Introduce multi mesh domains that are used  for multi mesh communication to reach and expose services. 
Make multi mesh envoy filter optional, as it could be better not to expose every service at once, but rather define specific service entries to do so one by one.

### Checklist

- [x] Implementation tested


Laci21 (approved):

pkg/resources/meshexpansion/multimesh.go
```diff
@@ -107,19 +117,19 @@ func (r *Reconciler) multimeshEnvoyFilter(selector map[string]string) *k8sutil.D
...
+func (r *Reconciler) multimeshDestinationRule(domain string) *k8sutil.DynamicObject {
 	return &k8sutil.DynamicObject{
 		Gvr: schema.GroupVersionResource{
 			Group:    "networking.istio.io",
 			Version:  "v1alpha3",
 			Resource: "destinationrules",
 		},
 		Kind:      "DestinationRule",
-		Name:      r.Config.WithRevision(multimeshResourceNamePrefix + "-destinationrule"),
+		Name:      r.Config.WithRevision(multimeshResourceNamePrefix + "-" + strings.ReplaceAll(domain, ".", "-")),
```
khernyo, line 128: My earlier suggestion to allow `-` in a domain name clashes with this. Multiple domains could be mapped to the same destinationrule name. So, `-` should either be kept as disallowed, or maybe use `GenerateName` here?

khernyo (approved):

pkg/resources/meshexpansion/multimesh.go
waynz0r: I changed it to use crc32 hash as suffix.

khernyo (approved):

Laci21 (approved):

PR #600, opened 2021-03-01
[1.8] support mixed network multi cluster

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

- implements support for setting network for remote clusters.
- use proper service ports on remote clusters 
- trigger `MeshGateway` reconcile on `Istio` CR change
- support for mesh expansion gateway using local endpoints only (needs Backyards Istio distribution for the actual feature)


pkg/apis/istio/v1beta1/remoteistio_types.go
```diff
@@ -80,6 +80,10 @@ type RemoteIstioSpec struct {
 	// Istio CR to which this gateway belongs to
 	IstioControlPlane *NamespacedName `json:"istioControlPlane,omitempty"`
 
+	// Network defines the network this cluster belong to. This name
```
Laci21, line 83:
```suggestion
	// Network defines the network this cluster belongs to. This name
```

pkg/controller/istio/istio_controller.go
```diff
@@ -477,14 +480,19 @@ func (r *ReconcileIstio) getMeshNetworks(config *istiov1beta1.Istio, logger logr
...
+		// to be backward compatible, the network name is the CR name if not specified
+		if networkName == "" {
+			networkName = remoteIstio.Name
+		}
+
+		if meshNetworks[networkName] == nil {
+			meshNetworks[networkName] = &istiov1beta1.MeshNetwork{}
 		}
+		meshNetworks[networkName].Endpoints = append(meshNetworks[networkName].Endpoints, istiov1beta1.MeshNetworkEndpoint{
+			FromRegistry: remoteIstio.Name,
```
Laci21, line 493: Might be missing something here, but shouldn't this be:
```suggestion
			FromRegistry: networkName,
```

pkg/resources/istiod/service.go
```diff
@@ -21,44 +21,49 @@ import (
...
 
+	"github.com/banzaicloud/istio-operator/pkg/apis/istio/v1beta1"
 	"github.com/banzaicloud/istio-operator/pkg/resources/templates"
 	"github.com/banzaicloud/istio-operator/pkg/util"
 )
 
+func ServicePorts(config *v1beta1.Istio) []apiv1.ServicePort {
+	return []apiv1.ServicePort{
+		{
+			// plaintext
```
Laci21, line 32: Nit: 
```suggestion
		// plaintext
```

pkg/controller/istio/istio_controller.go
waynz0r: The registry is the name of the cluster, not the network.

khernyo (approved):

pkg/controller/istio/istio_controller.go
Laci21: got it, thanks

Laci21 (approved):

PR #601, opened 2021-03-02
[1.8] Add custom tracing tags

### LINKED ISSUE #597: Add in support for custom_tags section in traces
**Is your feature request related to a problem? Please describe.**
I would like to be able to set a few custom_tags in our istio tracing environment. In my specific use case I would like to add a "cluster_name" tag to all traces inside that cluster. Istio will only write to the Jagger collector, so I cannot point to an agent for example and have that add the tag for me. 

**Describe the solution you'd like to see**
It seems like custom_tags is already supported in istio's config since 1.6.0 (https://github.com/istio/api/blob/1.6.0/mesh/v1alpha1/proxy.pb.go#L165)  and they have a doc page about it here: https://istio.io/v1.8/docs/tasks/observability/distributed-tracing/configurability/

**Describe alternatives you've considered**
I tried to see if I could point istio to an agent, but istio expects to send zipkin http requests not UDP/Thrift ones
I could build a custom proxy that accomplishes this but it would be a on-off custom proxy to handle functionality that exists in istio currently
I tried the annotation listed there, it works but I would have to update every app in our environment to include that annotation including data stores that we are using istio for mtls. Not impossible but a huge pain. 

**Additional context**
I was able to manually edit the config map this operator generates and manually add the custom_tag field and it worked just fine. So really this is just exposing that configuration via the Istio CRD.

nikore:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #597

### What's in this PR?
Adding in a new configuration option in the tracing section of the CRD to set custom tags to be added to all tracing spans that originate from istio (envoy proxy).

### Why?
In #597 I explain in greater detail but in short the envoy proxy only uses zipkin HTTP which is only ran on the collector. I want to point more than one cluster to the same collector so I would like to add a custom tag to specify which cluster the trace came from. Istio has supported this since 1.6 and seems to have been rather stable. 

### Additional context
I have been manually testing a build of this operator including my fix in my environments has been great thus far. I targeted this against 1.8 as that is what my environment uses currently, I am happy to try and port it over to 1.9 if needed.


pkg/apis/istio/v1beta1/istio_types.go
```diff
@@ -630,11 +631,12 @@ const (
...
-	Zipkin       ZipkinConfiguration       `json:"zipkin,omitempty"`
-	Lightstep    LightstepConfiguration    `json:"lightstep,omitempty"`
-	Datadog      DatadogConfiugration      `json:"datadog,omitempty"`
-	Strackdriver StrackdriverConfiguration `json:"stackdriver,omitempty"`
+	Tracer       TracerType                      `json:"tracer,omitempty"`
+	Zipkin       ZipkinConfiguration             `json:"zipkin,omitempty"`
+	Lightstep    LightstepConfiguration          `json:"lightstep,omitempty"`
+	Datadog      DatadogConfiugration            `json:"datadog,omitempty"`
+	Strackdriver StrackdriverConfiguration       `json:"stackdriver,omitempty"`
+	CustomTags   map[string]apiextensionsv1.JSON `json:"customTags,omitempty"`
```
Laci21, line 639: A short comment (one sentence) might be nice here to explain what these tags are for.

Laci21 (approved):
Thank you very much @nikore for your contribution, it looks good to me, I just added a small comment and please rebase to origin release-1.8 and regenerate the assets.

Also, once this merged it would be great if you could cherry-pick this to `release-1.9` as well as you mentioned.

nikore:
OK I added a comment to the field and created #608 to cover the 1.9 branch

pkg/apis/istio/v1beta1/istio_types.go
```diff
@@ -635,6 +636,8 @@ type TracingConfiguration struct {
 	Lightstep    LightstepConfiguration    `json:"lightstep,omitempty"`
 	Datadog      DatadogConfiugration      `json:"datadog,omitempty"`
 	Strackdriver StrackdriverConfiguration `json:"stackdriver,omitempty"`
+	//Custom tags to be added to spans, value can be literals, environment variables, or client requeset headers
```
Laci21, line 639:
```suggestion
	// Custom tags to be added to all spans, value can be literals, environment variables, or client request headers
```

Laci21 (approved):

khernyo (approved):

Laci21:
Thank you, @nikore! The feature has already landed in the `0.8.10` release.

PR #608, opened 2021-03-02
[1.9] Add custom tracing tags

### LINKED ISSUE #597: Add in support for custom_tags section in traces

nikore:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #597

### What's in this PR?
This is a cherry pick from this PR https://github.com/banzaicloud/istio-operator/pull/601 to the 1.9 branch


pkg/apis/istio/v1beta1/istio_types.go
```diff
@@ -648,6 +649,8 @@ type TracingConfiguration struct {
 	Lightstep    LightstepConfiguration    `json:"lightstep,omitempty"`
 	Datadog      DatadogConfiugration      `json:"datadog,omitempty"`
 	Strackdriver StrackdriverConfiguration `json:"stackdriver,omitempty"`
+	//Custom tags to be added to spans, value can be literals, environment variables, or client requeset headers
```
Laci21, line 652:
```suggestion
	// Custom tags to be added to all spans, value can be literals, environment variables, or client request headers
```

Laci21 (approved):

khernyo (approved):

PR #615, opened 2021-03-23
End-to-end test environment with two basic tests

khernyo:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | ASG-202

### What's in this PR?

A couple of small tweaks and an end-to-end test environment with two basic tests

The tests will be run for every push, but a failing end-to-end test will not block the merging of a PR.

### Checklist

- [x] Implementation tested


Makefile
```diff
@@ -116,13 +127,64 @@ verify-codegen: download-deps
...
+	docker pull kennethreitz/httpbin
+	kind load docker-image kennethreitz/httpbin
+
+	# TODO collect these from the operator (run the related functions and collect the referenced images)
+	docker pull gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0
+	docker pull docker.io/istio/pilot:${ISTIO_VERSION}
+	docker pull docker.io/istio/proxyv2:${ISTIO_VERSION}
+	kind load docker-image gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0
+	kind load docker-image docker.io/istio/pilot:${ISTIO_VERSION}
+	kind load docker-image docker.io/istio/proxyv2:${ISTIO_VERSION}
```
Laci21: Wouldn't this be nicer in a separate script file?

Makefile
```diff
@@ -116,13 +127,64 @@ verify-codegen: download-deps
...
+	kind load docker-image metallb/speaker:v0.9.6
+	./scripts/install-metallb.sh
+
+	docker pull kennethreitz/httpbin
+	kind load docker-image kennethreitz/httpbin
+
+	# TODO collect these from the operator (run the related functions and collect the referenced images)
+	docker pull gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0
+	docker pull docker.io/istio/pilot:${ISTIO_VERSION}
+	docker pull docker.io/istio/proxyv2:${ISTIO_VERSION}
```
Laci21: We probably should pull from gcr now.

Makefile
```diff
@@ -116,13 +127,64 @@ verify-codegen: download-deps
...
+	kind load docker-image ${IMG}
+	helm install --wait \
+		--set operator.image.repository=${IMAGE_REPOSITORY} \
+		--set operator.image.tag=${TAG} \
+		--create-namespace \
+		--namespace istio-system \
+		istio-operator-e2e-test \
+		deploy/charts/istio-operator/
+
+	# TODO maybe wait until all pods are up and running?
```
Laci21, line 173: The `--wait` flag should do this already:
```
--wait                         if set, will wait until all Pods, PVCs, Services, and minimum number of Pods of a Deployment, StatefulSet, or ReplicaSet are in a ready state before marking the release as successful. It will wait for as long as --timeout
```

test/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+1. set up a test env (creates a kind cluster with MetalLB and pre-loads docker images):
+
+       make e2e-test-env
+
+2. run the tests (builds the istio-operator docker image, installs it using helm and starts the tests):
+
+       make TAG=dev e2e-test
+
+The "TAG" variable is used for the istio-operator docker image tag. It can be anything, but you should
```
Laci21: If it can cause mixups shouldn't we randomize it maybe? Where does this tag matter at all?

test/e2e/istio_controller_suite_test.go
```diff
@@ -0,0 +1,74 @@
+/*
+Copyright 2019 Banzai Cloud.
```
Laci21:
```suggestion
Copyright 2021 Banzai Cloud.
```

test/e2e/testdata/TestIstioOperator/MeshGateway_sets_up_working_ingress.yaml
```diff
@@ -0,0 +1,114 @@
...
+    istio: mgw01
+  maxReplicas: 1
+  minReplicas: 1
+  replicaCount: 1
+  ports:
+    - name: http
+      port: 80
+      protocol: TCP
+      targetPort: 80
+  runAsRoot: true
```
Laci21: I wouldn't mind if we could test with this set to `false`, because that's the default and in that case there could be issues because of the less privileges.

test/e2e/testdata/TestIstioOperator/MeshGateway_sets_up_working_ingress.yaml
```diff
@@ -0,0 +1,114 @@
...
+        app: httpbin
+    spec:
+      terminationGracePeriodSeconds: 0
+      containers:
+        - image: kennethreitz/httpbin
+          imagePullPolicy: Always
+          name: httpbin
+          ports:
+            - containerPort: 80
+              protocol: TCP
```
Laci21: Where exactly do we indicate that this pod/namespace should be sidecar injected?

test/e2e/util/condition.go
```diff
@@ -0,0 +1,46 @@
...
+func WaitForCondition(timeout time.Duration, interval time.Duration, f func() (bool, error)) error {
+	start := time.Now()
+	timer := time.After(timeout)
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-timer:
+			return errors.Errorf("Timeout after %s", timeout)
```
Laci21:
```suggestion
			return errors.Errorf("timeout after %s", timeout)
```

test/e2e/istio_controller_test.go
```diff
@@ -0,0 +1,549 @@
+/*
+Copyright 2019 Banzai Cloud.
```
Laci21:
```suggestion
Copyright 2021 Banzai Cloud.
```

Laci21 (commented):
Might be just me but the tests in `test/e2e/istio_controller_test.go` seems a little hard to read at first glance. I don't have any good idea to make it better right now, but if you have an idea, feel free to go for it, or we can talk about this later as well (maybe Rod also has some ideas), doesn't need to address this concern in this PR.

It'd probably already help, if the file wasn't that long, so if you could move the helper functions (lister functions, etc.) to other files.

khernyo:
> Might be just me but the tests in `test/e2e/istio_controller_test.go` seems a little hard to read at first glance. I don't have any good idea to make it better right now, but if you have an idea, feel free to go for it, or we can talk about this later as well (maybe Rod also has some ideas), doesn't need to address this concern in this PR.
> 
> It'd probably already help, if the file wasn't that long, so if you could move the helper functions (lister functions, etc.) to other files.

Yes, I feel the same. I plan on improving it, but this will take time. We might switch to ginkgo, which would help a little with the structure, and I think I will just move the helpers into a separate file.

Makefile
khernyo: So, these must be the same as the images in `config/samples/istio_v1beta1_istio.yaml`. I think this would be an unrelated change and should be done in a separate PR.

Laci21: Okay, sure. It's fine for me btw until we don't hit issues with Docker rate limiting, but we should migrate at some point.

Makefile
khernyo: Yes and no: it only waits for the pods installed by helm. Usually, when the istio-operator is ready, a couple of pods in kube-system are still just starting up. It works out fine now, probably because there is a wait in `TestMain` for the cluster to be reachable. Waiting here for all pods might result in a lower load on the cluster when the actual tests start, so it might remove some flakiness.

I just added the above to the TODO.

Makefile
khernyo: fixed

test/README.md
khernyo: I wrote a short story about this. Hopefully, it makes more sense than the above.

test/e2e/testdata/TestIstioOperator/MeshGateway_sets_up_working_ingress.yaml
khernyo: done

test/e2e/testdata/TestIstioOperator/MeshGateway_sets_up_working_ingress.yaml
khernyo: Moved all these resources to a separate namespace with an `istio.io/rev` label. The httpbin pod is running with two containers now.

Laci21 (approved):
lgtm, thank you @khernyo!

waynz0r (approved):

PR #619, opened 2021-04-13
Fix missing RBAC permissions for Istio 1.8

turip:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

When trying to create a new Isito 1.8 based istio operator using these helm charts we are getting the following failure:

```
clusterroles.rbac.authorization.k8s.io "istio-reader-cp-v18x-istio-system" is forbidden: user "system:serviceaccount:istio-system:istio-operator-v18x" (groups=["system:serviceaccounts" "system:serviceaccounts:istio-system" "system:authenticated"]) is attempting to grant RBAC permissions not currently held:
{APIGroups:["authentication.istio.io"], Resources:["*"], Verbs:["get" "watch" "list"]}
{APIGroups:["config.istio.io"], Resources:["*"], Verbs:["get" "watch" "list"]}
```

This patch ensures that we are giving the istio-operator to provision 1.8 based Istio control planes too.


Laci21:
#620 should solve the root cause of this.

PR #628, opened 2021-04-20
[v2] Generate IstioControlPlane CRD resource and controller skeleton

Laci21:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### To Do

Tweaks will arrive in upcoming PRs to make it work properly.


Makefile
```diff
@@ -67,11 +67,11 @@ install-kubebuilder:
 
 # Install CRDs into a cluster
 install: manifests
-	kustomize build config/crd | kubectl apply -f -
+	bin/kustomize build config/crd | kubectl apply -f -
```
khernyo, line 70: Could you please add an item to your TODO list: pin tools to a specific version

khernyo (approved):

Makefile
Laci21: They are indeed pinned to a specific version: https://github.com/banzaicloud/istio-operator/blob/v2/Makefile#L10
Or do you mean something else?

khernyo: Sorry, this is perfect! I should've checked the install scripts...

PR #629, opened 2021-04-21
[v2] Bump kustomize version

Laci21:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

- Bump the used version of `kustomize`
- Chang the URL in the `kustomize` binary install script
- Fix a misused verb in a shell script comment

### Why?

- In the earlier `kustomize` versions the `namePrefix` field was used for `Namespace` kinds as well. This resulted in creating a `istio-operator-v2-istio-system` namespace instead of the desired `istio-system`. Related issue: https://github.com/kubernetes-sigs/kustomize/issues/235. This behaviour was changed for newer version - so that the `namePrefix` field is not prepended - but not for earlier versions, that's why I needed to update the version.
- The download location of `kustomize` was changed for newer versions, that's why the install script was modified as well.

### Checklist

- [x] Implementation tested


scripts/install_kubebuilder.sh
```diff
@@ -21,6 +21,6 @@ if [ ! -e bin/"${target_dir_name}" ]; then
     # download kubebuilder and extract it to tmp
     curl -L "https://go.kubebuilder.io/dl/${version}/${os}/${arch}" | tar -xz -C /tmp/
 
-    # extract the archive
+    # move the archive
     mv "/tmp/kubebuilder_${version}_${os}_${arch}" bin/"${target_dir_name}"
```
khernyo:
```suggestion
    # move the binary
    mv "/tmp/kubebuilder_${version}_${os}_${arch}" bin/"${target_dir_name}"
```
I would not be against removing this comment :)

khernyo (approved):

scripts/install_kubebuilder.sh
Laci21: You are right, deleted the comment :)

PR #631, opened 2021-04-21
[1.8] add mesh gateway service ip/hostname override annotations

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

Adds two annotations (service.banzaicloud.io/ip-address-override, service.banzaicloud.io/hostname-override) that can be used to override mesh gateway address in case the one in the status field of the service is different from the one needs to be used for external access.

### Why?

Most of the times the internal and external address of a service is different and while it would be the load balancer service controller's job to set proper external address to the status fields that is not always the case. So when necessary one can use the above annotations to specify the proper external address of a meshgateway.

Example:
```
spec:
  serviceAnnotations:
    service.banzaicloud.io/ip-address-override: 192.168.100.1
```

```
spec:
  serviceAnnotations:
    service.banzaicloud.io/hostname-override: lb-external-address.acme.org
```

### Checklist

- [x] Implementation tested

### DIFF
--- a/pkg/k8sutil/services.go
+++ b/pkg/k8sutil/services.go
@@ -25,8 +25,8 @@ import (
 )
 
 const (
-	ServiceIpAddressOverrideAnnotation = "service.banzaicloud.io/ip-address-override"
-	ServiceHostnameOverrideAnnotation  = "service.banzaicloud.io/hostname-override"
+	serviceIpAddressOverrideAnnotation = "service.banzaicloud.io/ip-address-override"
+	serviceHostnameOverrideAnnotation  = "service.banzaicloud.io/hostname-override"
 )
 
 type IngressSetupPendingError struct{}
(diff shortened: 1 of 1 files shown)


pkg/k8sutil/services.go
```diff
@@ -24,6 +24,11 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+const (
+	ServiceIpAddressOverrideAnnotation = "service.banzaicloud.io/ip-address-override"
+	ServiceHostnameOverrideAnnotation  = "service.banzaicloud.io/hostname-override"
```
Laci21: If they do not need to be exported to other files, then:
```suggestion
	serviceIpAddressOverrideAnnotation = "service.banzaicloud.io/ip-address-override"
	serviceHostnameOverrideAnnotation  = "service.banzaicloud.io/hostname-override"
```

pkg/k8sutil/services.go
```diff
@@ -33,40 +38,66 @@ func (e IngressSetupPendingError) Error() string {
 func GetServiceEndpointIPs(service corev1.Service) ([]string, bool, error) {
 	ips := make([]string, 0)
 
-	var hasHostname bool
+	var hostname string
+
+	if v, ok := service.GetAnnotations()[ServiceHostnameOverrideAnnotation]; ok && v != "" {
```
Laci21: Could `service.GetAnnotations()` return `nil`? Because in that case it should be handled too.

Laci21 (approved):
lgtm

khernyo (approved):
Awesome, thanks!

Issue #643, opened 2021-04-27
Istio-operator is not cleaned up properly on Kubernetes 1.20+


khernyo:
**Describe the bug**

When deleting an Istio resource on Kubernetes 1.20, some resources created by istio-operator are not deleted.
This is already a problem, but there is a more serious issue: one of the resources left in the cluster is a ValidatingWebhookConfiguration, and it prevents the reconciliation of any further Istio resources. Deleting the
leftover ValidatingWebhookConfiguration unblocks the reconciliation process which should now succeed.

There are three other resource types which are not cleaned up either: ClusterRole, ClusterRoleBinding and MutatingWebhookConfiguration. These don't cause any visible issues, though.

The cause is that in Kubernetes 1.20, ownership reference between cluster scoped and namespace scoped resources is disallowed, and such owner references are not followed by the GC.

For more information, see the note in section "Owners and Dependents" of https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents

**Steps to reproduce the issue:**

1. Create a cluster with Kubernetes 1.20
1. Install istio-operator (e.g. version 0.9.4)
1. List the ValidatingWebhookConfiguration resources. There should be only one, named "istio-operator-webhook"
1. Create an Istio resource
1. When the Istio resource is reconciled, delete it
1. List the ValidatingWebhookConfiguration resources again: There will be a new ValidatingWebhookConfiguration, named "istiod-...", which is never deleted

**Expected behavior**

No resources are left in the cluster inadvertently. E.g. leaving CRDs in the cluster is fine, but any resources which prevent
the correct functioning of the cluster is not.

**Additional context**

Kubernetes v1.20.2
Istio-operator v0.9.4

Laci21:
This is fixed in the latest version.

PR #644, opened 2021-04-27
Workaround for cluster cleanup issue in end-to-end tests

khernyo:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | #643 

### What's in this PR?

Several small end-to-end test fixes/cleanups, and a "workaround" for resource cleanup after deletion of Istio resources. The "workaround" is to use an older kubernetes version for end-to-end testing, which does clean up all resources.

### Checklist

- [x] Implementation tested

### DIFF
--- a/test/e2e/gvrs.go
+++ b/test/e2e/gvrs.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2021 Banzai Cloud.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import "k8s.io/apimachinery/pkg/runtime/schema"
+
+var serviceGVR = schema.GroupVersionResource{
+	Group:    "",
+	Version:  "v1",
+	Resource: "services",
+}
+
+var podGVR = schema.GroupVersionResource{
+	Group:    "",
+	Version:  "v1",
+	Resource: "pods",
+}
+
+var deploymentGVR = schema.GroupVersionResource{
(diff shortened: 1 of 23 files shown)


test/e2e/gvrs.go
```diff
@@ -0,0 +1,79 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import "k8s.io/apimachinery/pkg/runtime/schema"
+
+var serviceGVR = schema.GroupVersionResource{
```
Laci21: Aren't these defined somewhere already maybe?

Laci21 (approved):

test/e2e/gvrs.go
khernyo: I couldn't find any such definitions, but there are duplicated GVRs in istio-operator, so I'm gonna clean this up now

test/e2e/istio_controller_test_helpers.go
```diff
@@ -341,117 +332,65 @@ func HPAExists(ctx context.Context, t *testing.T, kubeClient client.Client, name
...
+		gvr.MutatingWebhookConfiguration,
+		gvr.DestinationRule,
+		gvr.VirtualService,
+		gvr.PeerAuthentication,
+		gvr.Gateway,
+		gvr.EnvoyFilter,
+		gvr.AttributeManifest,
+		gvr.IstioConfigHandler,
+		gvr.IstioConfigInstance,
+		gvr.IstioConfigRule,
```
Laci21: I think these are probably Mixer related resources, I don't think we need these anymore.

khernyo: Yeah, that's what I was thinking too. Shouldn't we just rip the mixer related code out?

Laci21: Actually, I wouldn't spend a minute on that. Mixer won't be in v2, that should be enough. I'd just delete these from here.

khernyo: Yeah, that makes sense.

Laci21 (approved):
lgtm

waynz0r (approved):

PR #657, opened 2021-05-11
V2 reconcile

martonsereg:
initial reconciliation logic based on operator-tools' helm reconciler


controllers/istiocontrolplane_controller.go
```diff
@@ -21,13 +21,27 @@ import (
...
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/discovery"
 	ctrl "sigs.k8s.io/controller-runtime"
+	ctrlBuilder "sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	istioclientv1alpha3 "github.com/banzaicloud/istio-client-go/pkg/networking/v1alpha3"
+	istioclientv1beta1 "github.com/banzaicloud/istio-client-go/pkg/security/v1beta1"
```
Laci21, line 39: We could consider starting to use the official Istio client in this v2 operator and see if it fits all our use cases and only move back to our client go if absolutely necessary.

deploy/charts/README-helm3.md
```diff
@@ -0,0 +1,66 @@
+# Helm v3 support
```
Laci21, line 1: Could we move all these charts to `assets` or `dependencies` or some other dir you think make sense?

internal/components/base/reconcile.go
```diff
@@ -0,0 +1,123 @@
+/*
```
Laci21, line 1: I'm wondering if it would make sense to have all these files in `pkg` instead of `internal`. I guess it would make sense at some point to reuse some of these functions/constants whatever externally, but I'm not sure, rather asking whether `internal` was a deliberate decision or we should at least consider `pkg`.

internal/components/base/reconcile.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/banzaicloud/istio-operator/v2/internal/static/gen/charts/base"
+	"github.com/banzaicloud/operator-tools/pkg/helm"
+	"github.com/banzaicloud/operator-tools/pkg/helm/templatereconciler"
+	"github.com/banzaicloud/operator-tools/pkg/types"
+)
+
+const (
+	ComponentName = "base"
+	ChartName     = "base"
+	ReleaseName   = "istio-operator-base"
```
Laci21: I think these all could be unexported variables at this point.
```suggestion
	componentName = "base"
	chartName     = "base"
	releaseName   = "istio-operator-base"
```

internal/static/main.go
```diff
@@ -0,0 +1,92 @@
...
+	})
+	if err != nil {
+		panic(fmt.Sprintf("failed to generate license vfs: %+v", err))
+	}
+
+	chartsPath := filepath.Join(getRepoRoot(), "deploy/charts")
+
+	for _, dir := range charts {
+		path := filepath.Join(chartsPath, dir)
+		// tmpPath := filepath.Join(filepath.Join(getRepoRoot(), "charts/tmp"), dir)
```
Laci21: left comment?

main.go
```diff
@@ -26,6 +26,9 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
+	istioclientv1alpha3 "github.com/banzaicloud/istio-client-go/pkg/networking/v1alpha3"
+	istioclientv1beta1 "github.com/banzaicloud/istio-client-go/pkg/security/v1beta1"
+
```
Laci21, line 31: same suggestion as before

internal/components/discovery/reconcile.go
```diff
@@ -0,0 +1,111 @@
...
+	istio_discovery "github.com/banzaicloud/istio-operator/v2/internal/static/gen/charts/istio-control/istio-discovery"
+	"github.com/banzaicloud/operator-tools/pkg/helm"
+	"github.com/banzaicloud/operator-tools/pkg/helm/templatereconciler"
+	"github.com/banzaicloud/operator-tools/pkg/types"
+)
+
+const (
+	ComponentName = "istio-discovery"
+	ChartName     = "istio-discovery"
+	ReleaseName   = "istio-operator-discovery"
```
Laci21: same
```suggestion
	componentName = "istio-discovery"
	chartName     = "istio-discovery"
	releaseName   = "istio-operator-discovery"
```

Laci21 (approved):

deploy/charts/README-helm3.md
martonsereg: yes, but tracked in another ticket

controllers/istiocontrolplane_controller.go
martonsereg: will open a ticket to investigate this

internal/components/base/reconcile.go
martonsereg: I can't really see reusing the reconciliation logic as is, so I'd say leave it in `internal` for now. Once we think that it should be reused, we should definitely move it.

PR #662, opened 2021-05-13
Initial add of Mixerless Telemetry Filter tests

spongeman66:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | Adds tests for Mixerless Telemetry Filters
| License         | Apache 2.0

### What's in this PR?
Adds tests for Mixerless Telemetry Filters

### Why?
Adds system level tests

### Additional context
This PR is not fully working yet. 

### Checklist

- [x] Implementation tested
- [x] Error handling code meets the [guideline](https://github.com/banzaicloud/pipeline/blob/master/docs/error-handling-guide.md)
- [x] Logging code meets the guideline
- [x] User guide and development docs updated (if needed)

### To Do

- [x] If the PR is not complete but you want to discuss the approach, list what remains to be done here

PR #663, opened 2021-05-14
Test Mixerless Telemetry filter adding and removal

spongeman66:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Adds End to End testing for Enabling and disabling Mixerless Telemetry Filters

### Why?
Adding E2E tests
An email was wrong in git, so this PR replaces 
https://github.com/banzaicloud/istio-operator/pull/662

### Checklist
- [x] Implementation tested

### DIFF
--- a/test/e2e/e2e_helpers_test.go
+++ b/test/e2e/e2e_helpers_test.go
@@ -45,7 +45,6 @@ import (
 	istiov1beta1 "github.com/banzaicloud/istio-operator/pkg/apis/istio/v1beta1"
 	"github.com/banzaicloud/istio-operator/pkg/k8sutil/mgw"
 	"github.com/banzaicloud/istio-operator/pkg/resources/gvr"
-	pkgutil "github.com/banzaicloud/istio-operator/pkg/util"
 	"github.com/banzaicloud/istio-operator/test/e2e/util"
 )
 
@@ -434,50 +433,33 @@ func GetIstioObject(istio *istiov1beta1.Istio, namespace, name string) error {
 		istio)
 }
 
-func GetMixerlessTelemetryStatus(istio *istiov1beta1.Istio, namespace, name string) (string, error) {
-
+func GetMixerlessTelemetryStatus(istio *istiov1beta1.Istio, namespace, name string) (*bool, error) {
 	err := GetIstioObject(istio, namespace, name)
 	if err != nil {
-		return "", err
+		return nil, err
 	}
 	// query current state and return it
-	if istio.Spec.MixerlessTelemetry.Enabled == nil {
-		return "N", nil
-	}
-	if *istio.Spec.MixerlessTelemetry.Enabled {
-		return "T", nil
-	}
-	return "F", nil
+	return istio.Spec.MixerlessTelemetry.Enabled, nil
 }
 
-func SetMixerlessTelemetryState(istio *istiov1beta1.Istio, namespace, name, newState string) (bool, error) {
-	var expectMissingFilter bool
-	err := GetIstioObject(istio, namespace, name)
-	if err != nil {
-		return false, err
-	}
-	switch newState {
-	case "T": // transition to true
-		istio.Spec.MixerlessTelemetry.Enabled = pkgutil.BoolPointer(true)
-		expectMissingFilter = false
-	case "F": // transition to false
-		istio.Spec.MixerlessTelemetry.Enabled = pkgutil.BoolPointer(false)
-		expectMissingFilter = true
-	case "N": // transition to nil
-		istio.Spec.MixerlessTelemetry.Enabled = nil
-		expectMissingFilter = true
-	}
-	return expectMissingFilter, nil
+func SetMixerlessTelemetryState(istio *istiov1beta1.Istio, newState *bool) error {
+	istio.Spec.MixerlessTelemetry.Enabled = newState
+	// upload to cluster
+	testEnv.Log.Info("Transitioning to:", "newState", newState)
+	return testEnv.Client.Update(context.TODO(), istio)
 }
 
 func WaitForMixerlessTelemetryFilter(
 	namespace, filterName string, expectMissing bool, timeout, interval time.Duration) error {
-
 	return util.WaitForCondition(timeout, interval, func() (bool, error) {
 		_, err := testEnv.Dynamic.Resource(gvr.EnvoyFilter).Namespace(namespace).Get(
 			context.TODO(),
 			filterName,
 			metav1.GetOptions{})
+		if err != nil && !k8sapierrors.IsNotFound(err) {
+			// only expect "Not Found" Error. Bail out here if we get an unexpected error
+			return false, err
+		}
 		if k8sapierrors.IsNotFound(err) == expectMissing {
 			// filter state matches expected
 			return true, nil
--- a/test/e2e/e2e_helpers_test.go
+++ b/test/e2e/e2e_helpers_test.go
@@ -45,6 +45,7 @@ import (
 	istiov1beta1 "github.com/banzaicloud/istio-operator/pkg/apis/istio/v1beta1"
 	"github.com/banzaicloud/istio-operator/pkg/k8sutil/mgw"
 	"github.com/banzaicloud/istio-operator/pkg/resources/gvr"
+	pkgutil "github.com/banzaicloud/istio-operator/pkg/util"
 	"github.com/banzaicloud/istio-operator/test/e2e/util"
 )
 
@@ -65,12 +66,10 @@ func maybeCleanup(log logr.Logger, noCleanupMsg string, cleanup func()) {
 }
 
 type IstioTestEnv struct {
-	log logr.Logger
-
-	c     client.Client
-	d     dynamic.Interface
-	istio *istiov1beta1.Istio
-
+	log                logr.Logger
+	c                  client.Client
+	d                  dynamic.Interface
+	istio              *istiov1beta1.Istio
 	clusterStateBefore ClusterResourceList
 }
 
@@ -426,4 +425,63 @@ func sortNamespacedNames(nns []types.NamespacedName) {
 func testDataPath(description ginkgo.GinkgoTestDescription) string {
 	path := filepath.Join(description.ComponentTexts...)
 	return strings.ReplaceAll(path, " ", "_")
+}
+
+func GetIstioObject(istio *istiov1beta1.Istio, namespace, name string) error {
+	return testEnv.Client.Get(context.TODO(), client.ObjectKey{
+		Namespace: namespace,
+		Name:      name},
+		istio)
+}
+
+func GetMixerlessTelemetryStatus(istio *istiov1beta1.Istio, namespace, name string) (string, error) {
+
+	err := GetIstioObject(istio, namespace, name)
+	if err != nil {
+		return "", err
+	}
+	// query current state and return it
+	if istio.Spec.MixerlessTelemetry.Enabled == nil {
+		return "N", nil
+	}
+	if *istio.Spec.MixerlessTelemetry.Enabled {
+		return "T", nil
+	}
+	return "F", nil
+}
+
+func SetMixerlessTelemetryState(istio *istiov1beta1.Istio, namespace, name, newState string) (bool, error) {
+	var expectMissingFilter bool
+	err := GetIstioObject(istio, namespace, name)
+	if err != nil {
+		return false, err
+	}
+	switch newState {
+	case "T": // transition to true
+		istio.Spec.MixerlessTelemetry.Enabled = pkgutil.BoolPointer(true)
+		expectMissingFilter = false
+	case "F": // transition to false
+		istio.Spec.MixerlessTelemetry.Enabled = pkgutil.BoolPointer(false)
+		expectMissingFilter = true
+	case "N": // transition to nil
+		istio.Spec.MixerlessTelemetry.Enabled = nil
+		expectMissingFilter = true
+	}
+	return expectMissingFilter, nil
+}
+
+func WaitForMixerlessTelemetryFilter(
+	namespace, filterName string, expectMissing bool, timeout, interval time.Duration) error {
+
+	return util.WaitForCondition(timeout, interval, func() (bool, error) {
+		_, err := testEnv.Dynamic.Resource(gvr.EnvoyFilter).Namespace(namespace).Get(
+			context.TODO(),
+			filterName,
+			metav1.GetOptions{})
+		if k8sapierrors.IsNotFound(err) == expectMissing {
+			// filter state matches expected
+			return true, nil
+		}
+		return false, nil
+	})
 }
--- a/test/e2e/e2e_helpers_test.go
+++ b/test/e2e/e2e_helpers_test.go
@@ -426,21 +426,21 @@ func testDataPath(description ginkgo.GinkgoTestDescription) string {
 	return strings.ReplaceAll(path, " ", "_")
 }
 
-func GetIstioObject(istio *istiov1beta1.Istio, namespace, name string) error {
+func getIstioObject(istio *istiov1beta1.Istio, namespace, name string) error {
 	return testEnv.Client.Get(context.TODO(), client.ObjectKey{
 		Namespace: namespace,
 		Name:      name},
 		istio)
 }
 
-func SetMixerlessTelemetryState(istio *istiov1beta1.Istio, newState *bool) error {
+func setMixerlessTelemetryState(istio *istiov1beta1.Istio, newState *bool) error {
 	istio.Spec.MixerlessTelemetry.Enabled = newState
 	// upload to cluster
 	testEnv.Log.Info("Updating cluster to:", "newState", newState)
 	return testEnv.Client.Update(context.TODO(), istio)
 }
 
-func WaitForMixerlessTelemetryFilter(
+func waitForMixerlessTelemetryFilter(
 	namespace, filterName string, filterShouldExist bool, timeout, interval time.Duration) error {
 	return util.WaitForCondition(timeout, interval, func() (bool, error) {
 		_, err := testEnv.Dynamic.Resource(gvr.EnvoyFilter).Namespace(namespace).Get(
@@ -460,12 +460,12 @@ func WaitForMixerlessTelemetryFilter(
 	})
 }
 
-func WaitForMixerlessTelemetryFilters(
+func waitForMixerlessTelemetryFilters(
 	namespace, filterName1 string, filterName2 string, filterShouldExist bool, timeout, interval time.Duration) error {
 	// check both filters sequentially. Usually, the first filter will wait the most.
-	err := WaitForMixerlessTelemetryFilter(namespace, filterName1, filterShouldExist, timeout, interval)
+	err := waitForMixerlessTelemetryFilter(namespace, filterName1, filterShouldExist, timeout, interval)
 	if err != nil {
 		return err
 	}
-	return WaitForMixerlessTelemetryFilter(namespace, filterName2, filterShouldExist, timeout, interval)
+	return waitForMixerlessTelemetryFilter(namespace, filterName2, filterShouldExist, timeout, interval)
 }
--- a/test/e2e/e2e_helpers_test.go
+++ b/test/e2e/e2e_helpers_test.go
@@ -433,15 +433,6 @@ func GetIstioObject(istio *istiov1beta1.Istio, namespace, name string) error {
 		istio)
 }
 
-func GetMixerlessTelemetryStatus(istio *istiov1beta1.Istio, namespace, name string) (*bool, error) {
-	err := GetIstioObject(istio, namespace, name)
-	if err != nil {
-		return nil, err
-	}
-	// query current state and return it
-	return istio.Spec.MixerlessTelemetry.Enabled, nil
-}
-
 func SetMixerlessTelemetryState(istio *istiov1beta1.Istio, newState *bool) error {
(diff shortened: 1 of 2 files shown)


test/e2e/e2e_test.go
```diff
@@ -87,6 +89,109 @@ var _ = Describe("E2E", func() {
...
+				verParts := strings.SplitN(string(version), ".", 3)
+				majMinor := fmt.Sprintf("%s.%s", verParts[0], verParts[1])
+				log.Info("Istio Version: ", "Version", majMinor)
+
+				// Names of the filters of interest
+				statsname = istio.WithRevision(fmt.Sprintf("mixerless-telemetry-stats-filter-%s", majMinor))
+				tcpname = istio.WithRevision(fmt.Sprintf("mixerless-telemetry-tcp-stats-filter-%s", majMinor))
+				log.Info("Filter Names: ", "Stats", statsname, "TCP", tcpname)
+
+			})
```
khernyo: All this should happen inside `It`. Please read https://onsi.github.io/ginkgo/#structuring-your-specs
Basically, the environment setup should happen in *BeforeEach`, but the test itself should be in `It`.

test/e2e/e2e_helpers_test.go
```diff
@@ -424,6 +423,65 @@ func sortNamespacedNames(nns []types.NamespacedName) {
...
+		return "", err
+	}
+	// query current state and return it
+	if istio.Spec.MixerlessTelemetry.Enabled == nil {
+		return "N", nil
+	}
+	if *istio.Spec.MixerlessTelemetry.Enabled {
+		return "T", nil
+	}
+	return "F", nil
```
khernyo: I don't think the "N", "T", "F" stuff helps. It should simply be:

```suggestion
	return istio.Spec.MixerlessTelemetry.Enabled, nil
```

test/e2e/e2e_test.go
```diff
@@ -87,6 +89,109 @@ var _ = Describe("E2E", func() {
...
+
+			AfterEach(func() {
```
khernyo: `AfterEach` should contain only teardown code, not a part of the test

test/e2e/e2e_helpers_test.go
```diff
@@ -424,6 +423,65 @@ func sortNamespacedNames(nns []types.NamespacedName) {
...
+		expectMissingFilter = false
+	case "F": // transition to false
+		istio.Spec.MixerlessTelemetry.Enabled = pkgutil.BoolPointer(false)
+		expectMissingFilter = true
+	case "N": // transition to nil
+		istio.Spec.MixerlessTelemetry.Enabled = nil
+		expectMissingFilter = true
+	}
+	return expectMissingFilter, nil
+}
```
khernyo: I have several issues with this function:

1. There is a `GetMixerlessTelemetryStatus` and a `SetMixerlessTelemetryStatus`. I would expect these to be symmetric, but they aren't. The "Set*" function first gets the object, overwriting any changes to the `istio` parameter, modifies the object and then it _doesn't_ update the object in the cluster. I would expect this function to only do the following: set the `Enabled` field and update the object on the cluster.
2. `expectMissingFilter` doesn't belong here.
Just a note: If you only look at the function signature, there is no way you can infer what that bool return value represents. In such cases it's useful to name the result values:
`func SetMixerlessTelemetryState(istio *istiov1beta1.Istio, namespace, name, newState string) (expectMissingFilter bool, err error) {`

3. the switch can be simplified if you replace the `newState string` parameter with `enabled *bool`. To set `expectMissingFilter`, you can use the "inverse" of `BoolPointer()`: `PointerToBool()`, which defaults to `false`.

test/e2e/e2e_helpers_test.go
```diff
@@ -424,6 +423,65 @@ func sortNamespacedNames(nns []types.NamespacedName) {
...
+	path := filepath.Join(description.ComponentTexts...)
+	return strings.ReplaceAll(path, " ", "_")
+}
+
+func GetIstioObject(istio *istiov1beta1.Istio, namespace, name string) error {
+	return testEnv.Client.Get(context.TODO(), client.ObjectKey{
+		Namespace: namespace,
+		Name:      name},
+		istio)
+}
```
khernyo: I think this would be cleaner:

```suggestion
func GetIstioObject(namespace, name string) (*istiov1beta1.Istio, error) {
	var istio istiov1beta1.Istio
	err := testEnv.Client.Get(context.TODO(), client.ObjectKey{
		Namespace: namespace,
		Name:      name},
		&istio)
	return &istio, err
}
```

(I would probably use `istiov1beta1.Istio` as the result type, but the pointer might come in handy on the caller side, so either is fine)

test/e2e/e2e_helpers_test.go
```diff
@@ -424,6 +423,65 @@ func sortNamespacedNames(nns []types.NamespacedName) {
...
+
+func WaitForMixerlessTelemetryFilter(
+	namespace, filterName string, expectMissing bool, timeout, interval time.Duration) error {
+
+	return util.WaitForCondition(timeout, interval, func() (bool, error) {
+		_, err := testEnv.Dynamic.Resource(gvr.EnvoyFilter).Namespace(namespace).Get(
+			context.TODO(),
+			filterName,
+			metav1.GetOptions{})
```
khernyo, line 449:
```suggestion
		_, err := testEnv.Dynamic.Resource(gvr.EnvoyFilter).Namespace(namespace).Get(
			context.TODO(),
			filterName,
			metav1.GetOptions{})
		if err != nil && !k8sapierrors.IsNotFound(err) {
			return false, err
		}
```

test/e2e/e2e_test.go
```diff
@@ -87,6 +89,109 @@ var _ = Describe("E2E", func() {
...
+
+			JustBeforeEach(func() {
+				// get the istio object created in the OUTER JustBeforeEach
+				log.Info("Namespace: ", "Namespace", instance.Namespace)
+				log.Info("Name: ", "Name", instance.Name)
+				filterBefore, err = GetMixerlessTelemetryStatus(&istio, instance.Namespace, instance.Name)
+				Expect(err).NotTo(HaveOccurred())
+				log.Info("JustBeforeEach: ", "filterBefore", filterBefore)
+				version := istio.Spec.Version // get first 2 digits
+				verParts := strings.SplitN(string(version), ".", 3)
```
khernyo: Nit:
```suggestion
				versionParts := strings.SplitN(string(version), ".", 3)
```

test/e2e/e2e_test.go
```diff
@@ -87,6 +89,109 @@ var _ = Describe("E2E", func() {
...
+				majMinor := fmt.Sprintf("%s.%s", verParts[0], verParts[1])
```
khernyo: Nit:
```suggestion
				majorMinor := fmt.Sprintf("%s.%s", verParts[0], verParts[1])
```

test/e2e/e2e_test.go
```diff
@@ -87,6 +89,109 @@ var _ = Describe("E2E", func() {
...
+				log.Info("Istio Version: ", "Version", majMinor)
+
+				// Names of the filters of interest
+				statsname = istio.WithRevision(fmt.Sprintf("mixerless-telemetry-stats-filter-%s", majMinor))
```
khernyo: Nit:
```suggestion
				statsName = istio.WithRevision(fmt.Sprintf("mixerless-telemetry-stats-filter-%s", majMinor))
```

test/e2e/e2e_test.go
```diff
@@ -87,6 +89,109 @@ var _ = Describe("E2E", func() {
...
+				tcpname = istio.WithRevision(fmt.Sprintf("mixerless-telemetry-tcp-stats-filter-%s", majMinor))
```
khernyo: Nit:
```suggestion
				tcpName = istio.WithRevision(fmt.Sprintf("mixerless-telemetry-tcp-stats-filter-%s", majMinor))
```

test/e2e/e2e_test.go
```diff
@@ -87,6 +89,109 @@ var _ = Describe("E2E", func() {
...
+						Expect(err).NotTo(HaveOccurred())
+
+						err = WaitForMixerlessTelemetryFilter(
+							istio.Namespace, tcpname, expectMissingFilter, 300*time.Second, 5*time.Second)
+						Expect(err).NotTo(HaveOccurred())
+					}
+				}
+			})
+
+			It("Transitions Filter through all states", func() {
```
khernyo: I think it would be better to avoid the `stateTransitions` variable and simply write down all steps. Also, it's enough to store the "enabled" state in the setup and reset it in teardown. The test itself doesn't have to depend on the initial state.

I'm thinking something like:

```
It("...", func() {
	SetMixerlessTelemetryState(istio, true)
	Expect(WaitForMixerlessTelemetryEnvoyFilters(istio, pkgutil.PointerToBool(true), 15*Seconds, 100*Millis)).To(Succeed())

	SetMixerlessTelemetryState(istio, false)
	Expect(WaitForMixerlessTelemetryEnvoyFilters(istio, pkgutil.PointerToBool(false), 15*Seconds, 100*Millis)).To(Succeed())

	SetMixerlessTelemetryState(istio, true)
	Expect(WaitForMixerlessTelemetryEnvoyFilters(istio, pkgutil.PointerToBool(true), 15*Seconds, 100*Millis)).To(Succeed())

	SetMixerlessTelemetryState(istio, nil)
	Expect(WaitForMixerlessTelemetryEnvoyFilters(istio, pkgutil.PointerToBool(nil), 15*Seconds, 100*Millis)).To(Succeed())
})
```

The signature of `WaitForMixerlessTelemetryEnvoyFilters` would be something like:
```
func WaitForMixerlessTelemetryEnvoyFilters(istio *Istio, shouldExist bool, timeout Duration, interval Duration) err
```

You can sprinkle it with a couple of `Sleep()` calls, although I don't think it's necessary here.

khernyo (changes requested):

test/e2e/e2e_test.go
spongeman66: done

test/e2e/e2e_test.go
spongeman66: done

test/e2e/e2e_test.go
spongeman66: done

test/e2e/e2e_test.go
spongeman66: done

test/e2e/e2e_test.go
spongeman66: No sleeps were needed since we poll for the change.

test/e2e/e2e_helpers_test.go
spongeman66: This method is called several times, and it seems problematic to create the istio instance inside the method, and then pass the address of that object outside the method.
It also seems wasteful to pass the entire object back each time this method is called.
Keeping this as-is

test/e2e/e2e_helpers_test.go
spongeman66: I think I have resolved all of these problems.
Get/Set MixerlessTelemetryStatus are more symmetrical and don't return the magic expectMissingFilter.
No switch is needed since we can directly set nil or use BoolPointer.

test/e2e/e2e_test.go
spongeman66: Ok, it only sets the final state.

test/e2e/e2e_helpers_test.go
spongeman66: This has been cleaned up without the Strings, and only uses a *bool

test/e2e/e2e_test.go
spongeman66: kept the initial filterBefore gathering and the majorMinor version gathering here, but moved the creation of the filter names into the test.

test/e2e/e2e_helpers_test.go
spongeman66: Good catch. Added checking for the unexpected error

test/e2e/e2e_test.go
spongeman66: modified to use filterShouldExist and added Expect().Should(Succeed())

test/e2e/e2e_helpers_test.go
khernyo: This is just a nit so I'm fine with leaving this as-is, but just to be clear:
- you can "stack allocate" a value and then return a pointer to it. I'm using quotes because there is no stack allocation in go. The compiler can allocate a value on the stack if it can prove that a pointer to that value doesn't escape the function, but this is just an optimization and it mustn't change the meaning of your program. If it does, then that's a compiler bug. So you are free to use any pointer any way you like.
- readable code is preferable to fast code, especially since most of the time is spent waiting for changes in the cluster.

test/e2e/e2e_test.go
spongeman66: Kept the setup stuff like filterBefore and majorMinor in the Before, but moved the specific filter names to the It()

test/e2e/e2e_test.go
spongeman66: It restores the beginning state only.

test/e2e/e2e_test.go
```diff
@@ -87,6 +89,83 @@ var _ = Describe("E2E", func() {
...
+				filterNow, err = GetMixerlessTelemetryStatus(&istio, instance.Namespace, instance.Name)
+				Expect(err).NotTo(HaveOccurred())
+				// Only change back if filterBefore != filterNow
+				// complicated by the ability of the *bool to be nil
+				if (filterNow == nil && filterBefore != nil) || (filterNow != nil && filterBefore == nil) || (
+					filterNow != nil && filterBefore != nil && *filterNow != *filterBefore) {
+					log.Info("AfterEach: Restore", "filterBefore", filterBefore, "filterNow", filterNow)
+					err = SetMixerlessTelemetryState(&istio, filterBefore)
+					Expect(err).NotTo(HaveOccurred())
+				}
```
khernyo: Is this really necessary? I mean `SetMixerlessTelemetryState(&istio, filterBefore)` should be enough, it doesn't matter what the current state is.

But even that's more than necessary: the istio resource will be deleted right after the test (look for `istioTestEnv.Close()`) and then the GC deletes all referencing resources, including the envoyfilters. What needs to be cleaned up, are the resources created "by hand", for example the meshgateway resource in one of the other test cases. There are no such resources in this test, so I think this `AfterEach` can be removed.

And I think that means the `filterBefore` variable is unnecessary, too.

spongeman66: Removed the tracking of the initial state and the cleanup

khernyo (approved):
Looks great, thanks @spongeman66!

waynz0r (approved):

test/e2e/e2e_helpers_test.go
```diff
@@ -424,6 +422,59 @@ func sortNamespacedNames(nns []types.NamespacedName) {
 }
 
 func testDataPath(description ginkgo.GinkgoTestDescription) string {
-       path := filepath.Join(description.ComponentTexts...)
-       return strings.ReplaceAll(path, " ", "_")
+	path := filepath.Join(description.ComponentTexts...)
+	return strings.ReplaceAll(path, " ", "_")
+}
+
+func GetIstioObject(istio *istiov1beta1.Istio, namespace, name string) error {
```
Laci21: Could this be an unexported function maybe?
```suggestion
func getIstioObject(istio *istiov1beta1.Istio, namespace, name string) error {
```

test/e2e/e2e_helpers_test.go
```diff
@@ -424,6 +422,59 @@ func sortNamespacedNames(nns []types.NamespacedName) {
...
+	return testEnv.Client.Get(context.TODO(), client.ObjectKey{
+		Namespace: namespace,
+		Name:      name},
+		istio)
+}
+
+func GetMixerlessTelemetryStatus(istio *istiov1beta1.Istio, namespace, name string) (*bool, error) {
```
Laci21: Where is this function used from?

test/e2e/e2e_helpers_test.go
```diff
@@ -424,6 +422,59 @@ func sortNamespacedNames(nns []types.NamespacedName) {
...
+	err := GetIstioObject(istio, namespace, name)
+	if err != nil {
+		return nil, err
+	}
+	// query current state and return it
+	return istio.Spec.MixerlessTelemetry.Enabled, nil
+}
+
+func SetMixerlessTelemetryState(istio *istiov1beta1.Istio, newState *bool) error {
```
Laci21: actually all functions in this file could be unexported imho
```suggestion
func setMixerlessTelemetryState(istio *istiov1beta1.Istio, newState *bool) error {
```

test/e2e/e2e_test.go
```diff
@@ -87,6 +89,64 @@ var _ = Describe("E2E", func() {
 				// TODO Check that the expected CRDs, deployments, services, etc. are present
 			})
 		})
+		Context("Mixerless Telemetry Stats Filter Test", func() {
+			// var resourcesFile string
```
Laci21: can this be deleted?

test/e2e/e2e_test.go
```diff
@@ -87,6 +89,64 @@ var _ = Describe("E2E", func() {
...
+				log.Info("Namespace: ", "Namespace", instance.Namespace)
+				log.Info("Name: ", "Name", instance.Name)
+				Expect(GetIstioObject(&istio, instance.Namespace, instance.Name)).Should(Succeed())
+				version := istio.Spec.Version // get first 2 digits
+				versionParts := strings.SplitN(string(version), ".", 3)
+				majorMinor = fmt.Sprintf("%s.%s", versionParts[0], versionParts[1])
+				log.Info("Istio Version: ", "Version", majorMinor)
+			})
+
+			It("Transitions Filter through all states", func() {
```
Laci21: I don't think that transitions actually matters in these tests, so it doesn't matter what was the previous state, it doesn't change the actual state.

Because of that, these could be written as three separate tests in my mind. When the enabled field is `true`, `false` and `nil`. Also, in my mind these should be separately runnable tests existing on their own, so they shouldn't be run after each other but be different small/reproducible test cases.

test/e2e/e2e_helpers_test.go
spongeman66: This method is used inside the test e2e_test.go, so it needs to be exported.

test/e2e/e2e_helpers_test.go
spongeman66: Good catch!
Since we don't query the status during setup anymore, this method is now unused.
I will remove it.

test/e2e/e2e_helpers_test.go
spongeman66: These are called from the test e2e_test.go, and need to be exported.

test/e2e/e2e_test.go
spongeman66: deleted

test/e2e/e2e_test.go
spongeman66: 1. The cluster is torn down and re-created between tests, so this would add 2-3 minutes per test, so we opted for only 1 test.
2. All states are tested, and the transitions between states are also tested. From a tester/customer perspective, a running cluster needs to be able to be changed without complete tear down and re-creation, so the transitions become important to the customer.

test/e2e/e2e_helpers_test.go
Laci21: As far as I know, since this function is used from the same package you don't need to export it.

test/e2e/e2e_helpers_test.go
Laci21: same as above for all similar functions in this file

test/e2e/e2e_test.go
Laci21: Of course, I agree with the first part, the cluster should NOT be torn down and recreated between test cases.

In my mind, there should be an initial setup, which brings the cluster to a state, which is a starting point for all test cases. Then each test case should do something and at the end clean up after itself so that the original starting point state is recreated and the next test case can go from there. (So not the whole cluster is torn down, only the initial state is recreated.)

So honestly, I'd do this with these mixerlesstelemetry test cases as well. I'm not convinced by your second point that it's worth testing the transitions this way. You also say that "a running cluster needs to be able to be changed without complete tear down and re-creation", which I agree with, but still these could be separate test cases in my mind as I suggested earlier without the complete tear down.

I might be wrong of course and I also don't want to hold this PR up even more, so if you insist on this solution it is fine by me.

spongeman66: So the transitions seem to do what you want.
Start with True.
Set to False and verify behavior. Restore cluster to true and verify behavior
Then set to nil and verify behavior. Restore cluster to true and verify behavior.
While these are not individual It() statements, they are separate steps within the same It()

The previous tests structured the before/after to build and teardown the cluster, so this test is impacted by that structure.

test/e2e/e2e_helpers_test.go
spongeman66: Fixed.
I am new to go, so I thought package == file.

test/e2e/e2e_test.go
spongeman66: Changed the comments in the test.
spreading out the states to individual tests increased run time by 2 minutes.

test/e2e/e2e_helpers_test.go
Laci21: No problem, thanks for the fix!

Laci21 (approved):
Thanks, @spongeman66!

PR #671, opened 2021-05-31
Update controller runtime to version 0.8.3

baluchicken:
This PR updates the controller runtime to version 0.8.3.
Also since it is a requirement it updates to go version from 1.14 to 1.16 with new alpine images.
In a nutshell:
- It uses the all new context driven controller which replaces the old channel based one.
- It uses the new client.Object


Laci21 (approved):
lgtm

pepov (approved):

PR #672, opened 2021-06-03
Istio test case to validate mesh gateway deployment and service are being created properly

shanchunyang0919:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | [SMM-190](https://jira-eng-sjc4.cisco.com/jira/browse/SMM-190)
| License         | Apache 2.0

## What's in this PR?
Implementation of test cases to validate: 
- A mesh gateway deployment is being created 
- A mesh gateway service is created that matches the mesh-gateway's name in the same namespace

## Why?
- To validate deployment's name matches the mesh-gateway's name and only istio-proxy sidecar container running inside
- To validate this mesh-gateway service is pointed to the mesh-gateway's deployment
- Make sure all resources being cleaned between each suites and test cases

### Environment Setup
- **KinD** cluster
- **istio-operator** deployed

### Logs and Pictures
- [Gist - Test cases logs](https://gist.github.com/shanchunyang0919/79affe052131a8521821df6f870142b8)
- **Clean up** after running test cases
Screen Shot 2021-06-02 at 8 48 01 PM

PR #676, opened 2021-06-04
Istio test case to validate mesh gateway and refactor e2e to fit BDD style

shanchunyang0919:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | [SMM-190](https://jira-eng-sjc4.cisco.com/jira/browse/SMM-190)

## **Note**
1. Refactored from this PR https://github.com/banzaicloud/istio-operator/pull/672

2. Make test cases to fit this format
```
Context("then the deployment")
- It (has at least 1 working pod)
Context("then the service")
- It has the given name
- It has the matching selectors
```

## What's in this PR?
Implementation of test cases to validate: 
- A mesh gateway deployment is being created 
- A mesh gateway service is created that matches the mesh-gateway's name in the same namespace

### Why?
- To validate deployment's name matches the mesh-gateway's name and only istio-proxy sidecar container running inside
- To validate this mesh-gateway service is pointed to the mesh-gateway's deployment
- Make sure all resources being cleaned between each suites and test cases

#### Environment Setup
- **KinD** cluster
- **istio-operator** deployed

#### Logs and Pictures
- [Gist - Test cases logs](https://gist.github.com/shanchunyang0919/235ee32eb4b4403ec50b3ae202ff276f)
- **Clean up** after running test cases
Screen Shot 2021-06-02 at 8 48 01 PM


test/e2e/e2e_test.go
```diff
@@ -148,35 +151,109 @@ var _ = Describe("E2E", func() {
...
+				})
+
+				It("should be pointed to the mesh gateway deployment", func() {
+					// Validate if mgw Service label selector matches mgw Deployment's MatchLabels field
+					Expect(mgwSvcLabelSelector).Should(BeEquivalentTo(mgwDeploymentMatchLabels))
+				})
+			})
+
+			Context("when mesh gateway address is created", func() {
+				// TODO: Add BeforeEach
```
turip: Please remove or fix this todo

test/e2e/e2e_helpers_test.go
```diff
@@ -421,10 +422,91 @@ func sortNamespacedNames(nns []types.NamespacedName) {
 	})
 }
 
+// Note: This method is currently not being used in the E2E test.
```
Laci21: I'm okay with deleting this entirely if not used from anywhere.

test/e2e/e2e_helpers_test.go
```diff
@@ -421,10 +422,91 @@ func sortNamespacedNames(nns []types.NamespacedName) {
...
+
+// Get unstructured object with Kuberentes dynamic clients.
+func GetUnstructuredObject(ctx context.Context, d dynamic.Interface, gvr schema.GroupVersionResource,
+	resource types.NamespacedName) (*unstructured.Unstructured, error) {
+	var (
+		unstructuredObject *unstructured.Unstructured
+		err                error
+	)
+
+	unstructuredObject, err = d.Resource(gvr).Namespace(resource.Namespace).Get(ctx, resource.Name, metav1.GetOptions{})
```
Laci21: As far as I know, it's more like the "GO way" to do this by:
```suggestion
	unstructuredObject, err := d.Resource(gvr).Namespace(resource.Namespace).Get(ctx, resource.Name, metav1.GetOptions{})
```

Same applies to all similar var declarations in this PR.

test/e2e/e2e_helpers_test.go
```diff
@@ -421,10 +422,91 @@ func sortNamespacedNames(nns []types.NamespacedName) {
...
+	return svc, nil
+}
+
+// Get a container list of given Deployment object.
+func GetContainersFromDeployment(dep *appsv1.Deployment) []corev1.Container {
+	return dep.Spec.Template.Spec.Containers
+}
+
+// Validate if the container exists in given container list.
+func ContainerExists(containerList []corev1.Container, containerName string) error {
```
Laci21: Please unexport the functions where possible:
```suggestion
func containerExists(containerList []corev1.Container, containerName string) error {
```

test/e2e/e2e_test.go
shanchunyang0919: Thanks for the comment Peter. Just addressed it by removing `TODO`.

There is no really a practical use for us to add BeforeEach to there now.  
It is more of future usages when we could maybe add more test cases on that `"when mesh gateway address is created"` block.

Laci21 (approved):
LGTM, @shanchunyang0919, thanks for your PR!

I've only added a few nits.

test/e2e/e2e_helpers_test.go
shanchunyang0919: > I'm okay with deleting this entirely if not used from anywhere.

Thank you for the comment. Addressed it by removing it entirely.

test/e2e/e2e_helpers_test.go
shanchunyang0919: Thanks for the comment. Addressed it.

test/e2e/e2e_helpers_test.go
shanchunyang0919: Thanks for the comment. Addressed it by make those functions un-exported.

PR #681, opened 2021-06-15
[v2] Dynamically register Istio CRD watches

Laci21:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

1. Add some missing RBAC rules for the `IstioControlPlane` controller
2. Setup Istio CRD watches in the reconcile function (but only once) after the Istio CRDs are installed

### Why?

1. Controller failed without the RBAC rules
2. Controller failed when some Istio CRDs were not installed and tried to be watched

### Additional context

Here's how the watch error looked like:
```
2021-06-15T10:27:00.924+0200	ERROR	controller-runtime.source	if kind is a CRD, it should be installed before calling Start	{"kind": "EnvoyFilter.networking.istio.io", "error": "no matches for kind \"EnvoyFilter\" in version \"networking.istio.io/v1alpha3\""}
```

### Checklist

- [x] Implementation tested

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -10,6 +10,7 @@ rules:
   - ""
   resources:
   - configmaps
+  - endpoints
   - secrets
   - serviceaccounts
   - services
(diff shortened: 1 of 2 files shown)


martonsereg (approved):

martonsereg:
thanks @Laci21

waynz0r (approved):

PR #686, opened 2021-06-18
MeshGatway API protos

martonsereg:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| API breaks?     | yes
| Deprecations?   | no
| Related tickets | fixes https://jira-eng-sjc4.cisco.com/jira/browse/SMM-183

### What's in this PR?
MeshGateway API protos and generated go code

### TODO
- [x] fix build
- [x] add comments
- [x] figure out if we can generate pointers for primitive types


Makefile
```diff
@@ -115,7 +115,7 @@ endif
 generate: download-deps update-istio-deps
 	cd build && ../bin/buf generate --path api
 	cd api/v1alpha1 && ../../bin/controller-gen object:headerFile="../../hack/boilerplate.go.txt" paths="./..."
-	go run static/generate.go
+	cd static && go run generate.go
```
Laci21, line 118: Out of curiosity: Is it better like this for some reason? :)

api/v1alpha1/common.proto
```diff
@@ -0,0 +1,76 @@
...
+
+package istio_operator.v2.api.v1alpha1;
+
+option go_package = "github.com/banzaicloud/istio-operator/v2/api/v1alpha1";
+
+message NamespacedName {
+    // Name of the referenced Kubernetes resource
+    string name = 1 [(gogoproto.jsontag) = "name,omitempty"];
+    // Namespace of the referenced Kubernetes resource
+    string namespace = 2  [(gogoproto.jsontag) = "namespace,omitempty"];;
```
Laci21: Nit:
```suggestion
    string namespace = 2  [(gogoproto.jsontag) = "namespace,omitempty"];
```

api/v1alpha1/common.proto
```diff
@@ -0,0 +1,76 @@
...
+}
+
+message GatewaySDSConfiguration {
+    // Flags if Gateway Secret Discovery Service is enabled or not
+    google.protobuf.BoolValue Enabled = 1 [(gogoproto.wktpointer) = true, (gogoproto.jsontag) = "enabled,omitempty"];
+    // Container image to use for SDS
+    string Image = 2 [(gogoproto.jsontag) = "image,omitempty"];
+    // Standard Kubernetes memory and CPU resource requirements
+    k8s.io.api.core.v1.ResourceRequirements Resources = 3 [(gogoproto.jsontag) = "resources,omitempty"];
+}
```
Laci21: I might remember incorrectly, but aren't these settings for the `node-agent` component, which was deprecated? If so, then we probably won't need these settings at all. @waynz0r, could you help out in this and confirm or deny what I wrote? :)

(Also, even if we should keep this config, imho this should go to the meshgateway spec instead of the common spec.)

api/v1alpha1/meshgateway.proto
```diff
@@ -0,0 +1,119 @@
...
+// -->
+//
+// <!-- go code generation tags
+// +genclient
+// +k8s:deepcopy-gen=true
+// -->
+message MeshGatewaySpec {
+    // Contains the intended version for the Mesh Gateway.
+    // +kubebuilder:validation:Pattern=^1.
+    string version = 1 [(gogoproto.jsontag) = "version", (google.api.field_behavior) = REQUIRED];
```
Laci21: Why should we have a `version` for the MGW resource, what would be the purpose of that? A gateway could be bound to any Istio control plane, which can be set below in the config. During an upgrade e.g. to user should change the version too to the upgraded version? Or should it be left intact? In that case it might be confusing that the version is different than the corresponding istio control plane version.

api/v1alpha1/meshgateway.proto
```diff
@@ -0,0 +1,119 @@
...
+    MeshGatewayConfiguration MeshGwConfiguration = 2 [(gogoproto.nullable) = false, (gogoproto.embed) = true, (gogoproto.jsontag) = ",inline"];
+    // Service ports to open for the gateway
+    // +kubebuilder:validation:MinItems=0
+    repeated ServicePort ports = 3 [(gogoproto.nullable) = false, (gogoproto.jsontag) = "ports"];
+    // Type of gateway, either ingress or egress
+    // +kubebuilder:validation:Enum=ingress;egress
+    GatewayType type = 4  [(gogoproto.jsontag) = "type"];
+    // Istio CR to which this gateway belongs to
+    NamespacedName istioControlPlane = 5 [(gogoproto.jsontag) = "istioControlPlane,omitempty"];
+}
```
Laci21: Are these fields generated as required fields like this? I guess not, so I'd suggest using the necessary annotation for that like this:

```suggestion
    repeated ServicePort ports = 3 [(gogoproto.nullable) = false, (gogoproto.jsontag) = "ports"], (google.api.field_behavior) = REQUIRED];
    // Type of gateway, either ingress or egress
    // +kubebuilder:validation:Enum=ingress;egress
    GatewayType type = 4  [(gogoproto.jsontag) = "type"], (google.api.field_behavior) = REQUIRED];
    // Istio CR to which this gateway belongs to
    NamespacedName istioControlPlane = 5 [(gogoproto.jsontag) = "istioControlPlane"], (google.api.field_behavior) = REQUIRED];
}
```

The `istioControlPlane` field was not a required field in the v1 version, but I think that was only for backward compatibility reasons, imho we can simply make it required as well.

Also, the same applies to all other required fields.

api/v1alpha1/meshgateway.proto
```diff
@@ -0,0 +1,119 @@
...
+
+message MeshGatewayConfiguration {
+    // Standard Kubernetes resource configuration
+    BaseK8sResourceConfigurationWithHPAWithoutImage BaseK8sResourceConfigWithHPAWithoutImage = 1 [(gogoproto.nullable) = false, (gogoproto.embed) = true, (gogoproto.jsontag) = ",inline"];
+    // Labels to attach to the mesh gateway
+	map<string, string> Labels = 2 [(gogoproto.jsontag) = "labels,omitempty"];
```
Laci21: These labels were always misleading to the end users where to set these, because labels are usually set in every resource outside of the spec and this field is at `spec.labels`. It's fine like this in this PR, I'd suggest to either create a ticket for this or let's just discuss how to solve this better.

api/v1alpha1/meshgateway.proto
```diff
@@ -0,0 +1,119 @@
...
+    // Reconciliation status of the mesh gateway
+    ConfigState Status = 1 [(gogoproto.jsontag) = "Status,omitempty"];
+	// Current address for the gateway
+    repeated string GatewayAddress = 2 [(gogoproto.jsontag) = "GatewayAddress,omitempty"];
+    // Reconciliation error message if any
+    string ErrorMessage = 3 [(gogoproto.jsontag) = "ErrorMessage,omitempty"];
+}
+
+enum ConfigState {
+    Created = 0;
```
Laci21, line 109: identation?

config/samples/servicemesh_v1alpha1_meshgateway.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: servicemesh.cisco.com/v1alpha1
+kind: MeshGateway
+metadata:
+  name: meshgateway-sample
+spec:
```
Laci21, line 5: We could add an example spec here to be able to try it easily.

Makefile
martonsereg: the static folder is a separate gomodule, so you'll want to run `go run` from that folder because of dependencies.

martonsereg: actually, it wasn't even working from me without these changes, because the vfsgen dependencies were missing

Laci21: I see, thanks!

api/v1alpha1/common.proto
martonsereg: you're probably right, I was blindly copying fields from the previous CR and missed it. I'll delete it.

api/v1alpha1/meshgateway.proto
martonsereg: What do you mean by generated as required fields? I don't think the `REQUIRED` tag is really making a difference here, but I can add them.

api/v1alpha1/meshgateway.proto
martonsereg: seems good to me 😄

api/v1alpha1/meshgateway.proto
martonsereg: It's a copy-paste error, thanks for noticing :)

api/v1alpha1/meshgateway.proto
Laci21: It should make a difference for the CRD generation: https://github.com/banzaicloud/istio-operator/blob/v2/config/crd/bases/istio-operator-crds.gen.yaml#L1490-L1491

api/v1alpha1/meshgateway.proto
Laci21: the issue is tabs vs spaces :)

api/v1alpha1/meshgateway.proto
martonsereg: I'd leave it as is, until we figure out something better. But we can return to the question when doing the reconciliation.

api/v1alpha1/meshgateway.proto
Laci21: and it's here: https://github.com/banzaicloud/istio-operator/pull/686/commits/20fcca190848a9c27287de1325c63d00299a9c69#diff-b3f87a119ff83e86985bcac0532f13a79e219aac37a9879c0c8b024e367082eeR2356-R2358

api/v1alpha1/meshgateway.proto
Laci21: Sure, I'm not holding this PR up, but please let's get back to this because I know for sure that this was misleading multiple times for multiple people (once for me as well :) ).

Laci21 (approved):
lgtm, thanks @martonsereg!

PR #690, opened 2021-07-02
[v2] refactor MGW api

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        |no
| New feature?    |no
| API breaks?     | yes
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

Contains MGW api refactor to get rid of embedded structs.

### Why?

Unfortunately embedded structs cannot be fully supported with protobuf and causes more issues than it solves.


martonsereg (commented):
could you please update the meshgateway sample to match the new structure?

api/v1alpha1/meshgateway.proto
```diff
@@ -50,46 +50,55 @@ option go_package = "github.com/banzaicloud/istio-operator/v2/api/v1alpha1";
...
+
+    GroupVersionKind groupVersionKind = 1 [(gogoproto.nullable) = false];
+    NamespacedName objectKey = 2;
+    repeated Patch patches = 3 [(gogoproto.nullable) = false];
 }
 
 enum GatewayType {
-    Ingress = 0;
-    Egress = 1;
+    unspecified = 0;
```
Laci21, line 99: Why are these `unspecified`s needed in the Enums?

Laci21 (approved):
lgtm

api/v1alpha1/meshgateway.proto
waynz0r: unspecified was introduced because the 0 value is not represented in marshalled json or yaml.

PR #695, opened 2021-07-12
[v1.10] fix nil annotations in injection template

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

The injection template contains a bug that removes pod annotations in case of multiple containers in a pod.

### Why?

In case of multiple containers the parsed template contains a nil annotation instead of an empty set.

### Checklist

- [x] Implementation tested


martonsereg (approved):

Laci21 (approved):
Thank you, @waynz0r!

PR #702, opened 2021-07-15
[v1.10] k8s resource overlay patches

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

Adds support to set overlay patches for k8s resources in the Istio CR.

### Why?

Overlay patches provide a way to modify almost any part of the k8s resources before apply, thus give a way the override settings or add changes which cannot be modified through the CR directly.

### Example

```
spec:
  k8sResourceOverlays:
  - groupVersionKind:
      group: apps
      kind: Deployment
      version: v1
    objectKey:
      name: istiod-cp-v19x
    patches:
    - parseValue: false
      path: /metadata/labels?/patched-label
      type: replace
      value: patched-value
    - parseValue: true
      path: /spec/template/spec/volumes/-
      type: replace
      value: |
        name: external-ca-cert
        secret:
          secretName: external-ca-cert
          optional: true
    - parseValue: true
      path: /spec/template/spec/containers/name=discovery/volumeMounts/-
      type: replace
      value: |
        name: external-ca-cert
        mountPath: /etc/external-ca-cert
        readOnly: true
```


Laci21 (approved):
lgtm

PR #713, opened 2021-08-17
[v2] mesh gateway reconciliation and various other changes

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

Basic mesh gateway reconciliation added.


api/v1alpha1/common.proto
```diff
@@ -115,8 +117,43 @@ message BaseKubernetesResourceConfig {
     // Standard Kubernetes pod annotation and label configuration
     K8sObjectMeta podMetadata = 15;
 
-    // K8s resource overlay patches
-    repeated K8sResourceOverlayPatch resourceOverlays = 16;
+    // PodDisruptionBudget configuration
+    PodDisruptionBudget podDisruptionBudget = 16;
+
+    DeploymentStrategy deploymentStrategy = 17;
```
Laci21, line 123: we should make sure these are used in the istio-discovery helm chart as well, so that all `icp.spec.istiod.deployment` CR fields have actual effect on the charts.

Laci21: BTW, these types are not declared by chance in some k8s protos already?

api/v1alpha1/common.proto
```diff
@@ -115,8 +117,43 @@ message BaseKubernetesResourceConfig {
...
+    // +optional
+    string type = 1;
+
+    // Rolling update config params. Present only if DeploymentStrategyType =
+    // RollingUpdate.
+    // +optional
+    RollingUpdateDeployment rollingUpdate = 2;
+}
+
+// is a description of a PodDisruptionBudget
```
Laci21: half comment is missing here maybe?

config/samples/servicemesh_v1alpha1_meshgateway.yaml
```diff
@@ -1,45 +1,113 @@
 apiVersion: servicemesh.cisco.com/v1alpha1
 kind: MeshGateway
 metadata:
-  name: meshgateway-sample
+  name: joskapista
```
Laci21:
```suggestion
  name: mgw-sample
```

:)

config/samples/servicemesh_v1alpha1_meshgateway.yaml
```diff
@@ -1,45 +1,113 @@
...
+        configMap:
+          localObjectReference:
+            name: log-config
+          items:
+            - key: log_level
+              path: log_level
   istioControlPlane:
-      name: cp-v110x
-      namespace: istio-system
+    name: cp-v19x
```
Laci21:
```suggestion
    name: cp-v110x
```

config/samples/servicemesh_v1alpha1_meshgateway.yaml
```diff
@@ -1,45 +1,113 @@
...
+    namespace: istio-system
+  k8sResourceOverlays:
+  - groupVersionKind:
+      kind: DestinationRule
+    objectKey:
+      name: meshexpansion-dr-istiod-cp-v19x
```
Laci21:
```suggestion
      name: meshexpansion-dr-istiod-cp-v110x
```

controllers/common.go
```diff
@@ -0,0 +1,82 @@
...
+		err := c.Get(ctx, key, icp)
+		if err == nil {
+			return icp, nil
+		}
+		if err != nil {
+			return nil, errors.WrapIf(err, "could not get related Istio control plane")
+		}
+	}
+
+	// get the oldest otherwise for backward compatibility
```
Laci21: backward compatibility with what?

controllers/common.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil, errors.WrapIf(err, "could not list istio control planes")
+	}
+	if len(icps.Items) == 0 {
+		return nil, errors.New("no Istio control planes were found")
+	}
+
+	sort.Sort(v1alpha1.SortableIstioControlPlaneItems(icps.Items))
+
+	icp = &icps.Items[0]
+	icp.SetGroupVersionKind(v1alpha1.SchemeBuilder.GroupVersion.WithKind("IstioControlPlane"))
```
Laci21: Why is this needed btw?

controllers/meshgateway_controller.go
```diff
@@ -39,15 +55,161 @@ type MeshGatewayReconciler struct {
...
+	// do not reqeue if control plane was not found
+	if icp == nil && err == nil {
+		return ctrl.Result{
+			Requeue: false,
+		}, nil
+	}
+
+	if !isIstioVersionSupported(icp.Spec.Version) {
+		return ctrl.Result{}, nil
+	}
```
Laci21, line 81: Shouldn't we put `requeue: false` on the result and also log an error maybe?
this is from icp reconciler:
```suggestion
	if !isIstioVersionSupported(icp.Spec.Version) {
		err = errors.New("intended Istio version is unsupported by this version of the operator")
		logger.Error(err, "", "version", icp.Spec.Version)

		return reconcile.Result{
			Requeue: false,
		}, nil
	}
```

controllers/meshgateway_controller.go
```diff
@@ -39,15 +55,161 @@ type MeshGatewayReconciler struct {
...
+
+func (r *MeshGatewayReconciler) getRelatedIstioControlPlane(ctx context.Context, c client.Client, mgw *servicemeshv1alpha1.MeshGateway, logger logr.Logger) (*servicemeshv1alpha1.IstioControlPlane, error) {
+	icp, err := GetRelatedIstioCR(context.Background(), r.GetClient(), client.ObjectKey{
+		Name:      mgw.GetSpec().GetIstioControlPlane().GetName(),
+		Namespace: mgw.GetSpec().GetIstioControlPlane().GetNamespace(),
+	})
+	if err != nil {
+		updateErr := components.UpdateStatus(ctx, c, mgw, components.ConvertConfigStateToReconcileStatus(servicemeshv1alpha1.ConfigState_ReconcileFailed), err.Error())
+		if updateErr != nil {
+			logger.Error(updateErr, "failed to update state")
```
Laci21: same suggestion for all similar error logs
```suggestion
			logger.Error(updateErr, "failed to update mesh gateway state")
```

controllers/meshgateway_controller.go
```diff
@@ -39,15 +55,161 @@ type MeshGatewayReconciler struct {
...
+
+			return nil, errors.WithStack(err)
+		}
+		if err != nil {
+			logger.Error(err, "could not get related istio control plane")
+		}
```
Laci21: Is this needed here? `err` seems to be always not nil

go.mod
```diff
@@ -28,4 +48,7 @@ require (
 replace (
 	github.com/banzaicloud/istio-operator/v2/api => ./api
 	github.com/banzaicloud/istio-operator/v2/static => ./static
+
+	// needs a fork to support int64/uint64 marshalling to integers
+	github.com/gogo/protobuf => github.com/waynz0r/protobuf v1.3.3-0.20210811122234-64636cae0910
```
Laci21, line 53: have you considered contributing this upstream?

internal/assets/manifests/meshgateway/templates/deployment.yaml
```diff
@@ -0,0 +1,77 @@
...
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ $gateway.name }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+{{- include "deployment.labels" . | indent 4 }}
+{{- include "toYamlIf" (dict "value" $gateway.metadata.annotations "key" "annotations" "indent" 2) | indent 2 }}
+spec:
+{{- if not $gateway.autoscaleEnabled }}
```
Laci21: I guess this should be removed

internal/assets/manifests/meshgateway/templates/service.yaml
```diff
@@ -0,0 +1,35 @@
...
+spec:
+{{- if $service.loadBalancerIP }}
+  loadBalancerIP: "{{ $service.loadBalancerIP }}"
+{{- end }}
+{{- if $service.loadBalancerSourceRanges }}
+  loadBalancerSourceRanges:
+{{ toYaml $service.loadBalancerSourceRanges | indent 4 }}
+{{- end }}
+{{- if $service.externalTrafficPolicy }}
+  externalTrafficPolicy: {{$service.externalTrafficPolicy }}
```
Laci21:
```suggestion
  externalTrafficPolicy: {{ $service.externalTrafficPolicy }}
```

internal/assets/manifests/meshgateway/values.yaml.tpl
```diff
@@ -0,0 +1,53 @@
...
+{{ toYamlIf (dict "value" .GetPodDisruptionBudget "key" "podDisruptionBudget") | indent 2 }}
+{{ toYamlIf (dict "value" .GetPodMetadata "key" "podMetadata") | indent 2 }}
+{{- end }}
+
+{{- with .GetSpec.GetService }}
+service:
+  type: {{ .GetType }}
+{{ toYamlIf (dict "value" .GetMetadata "key" "metadata") | indent 2 }}
+{{ toYamlIf (dict "value" .GetPorts "key" "ports") | indent 2 }}
+{{ toYamlIf (dict "value" .GetPorts "key" "ports") | indent 2 }}
```
Laci21:
```suggestion
{{ toYamlIf (dict "value" .GetPorts "key" "ports") | indent 2 }}
```

internal/components/meshgateway/meshgateway_test.go
```diff
@@ -0,0 +1,137 @@
...
+	var expectedValues map[string]interface{}
+	if err := yaml.Unmarshal(mgwExpectedValues, &expectedValues); err != nil {
+		t.Fatal(err)
+	}
+
+	if !reflect.DeepEqual(values, expectedValues) {
+		valuesYaml, err := yaml.Marshal(values)
+		if err != nil {
+			t.Fatal(err)
+		}
```
Laci21: Is the yaml.Unmarshal + reflect.DeepEqual + yaml.Marshal needed here? Can't you just call the `if len(report.Diffs) > 0 {` after `util.CompareYAMLs(mgwExpectedValues, valuesYaml)`? Similarly how it's done in `TestMGWResourceDump`.

internal/components/meshgateway/reconcile.go
```diff
@@ -0,0 +1,172 @@
...
+func (rec *Reconciler) values(object runtime.Object) (helm.Strimap, error) {
+	mgw, ok := object.(*v1alpha1.MeshGateway)
+	if !ok {
+		return nil, errors.WrapIff(errors.NewPlain("object cannot be converted to a MeshGateway"), "%+v", object)
+	}
+
+	values, err := util.TransformICPToStriMapWithTemplate(&v1alpha1.MeshGatewayWithProperties{
+		MeshGateway: mgw,
+		Properties: v1alpha1.MeshGatewayProperties{
+			Revision:              "cp-v110x.istio-system",
```
Laci21: Is this okay to be static?

internal/util/testdata/test_values.yaml.tmpl
```diff
@@ -2,7 +2,7 @@
 
 ## Discovery Settings
 pilot:
-  autoscaleEnabled: {{ or .GetSpec.GetIstiod.GetDeployment.GetReplicas.GetAutoscaleEnabled true }}
+  autoscaleEnabled: false
```
Laci21:
```suggestion
  autoscaleEnabled: {{ and (gt (.GetSpec.GetIstiod.GetDeployment.GetReplicas.GetMin | int) 0) (gt (.GetSpec.GetIstiod.GetDeployment.GetReplicas.GetMax | int) (.GetSpec.GetIstiod.GetDeployment.GetReplicas.GetMin | int)) }}
```

internal/util/util.go
```diff
@@ -18,23 +18,29 @@ package util
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
 
-	"github.com/banzaicloud/istio-operator/v2/api/v1alpha1"
 	"github.com/banzaicloud/operator-tools/pkg/helm"
 	"github.com/banzaicloud/operator-tools/pkg/utils"
 )
 
-func TransformICPToStriMapWithTemplate(icp *v1alpha1.IstioControlPlane, filesystem fs.FS, templateFileName string) (helm.Strimap, error) {
+func TransformICPToStriMapWithTemplate(data interface{}, filesystem fs.FS, templateFileName string) (helm.Strimap, error) {
```
Laci21:
```suggestion
func TransformStructToStriMapWithTemplate(data interface{}, filesystem fs.FS, templateFileName string) (helm.Strimap, error) {
```

api/v1alpha1/common.proto
waynz0r: They are, but we need custom intorstring types unfortunately.

controllers/common.go
waynz0r: we can get rid of that i just copied the functionality from the v1 branch.

controllers/meshgateway_controller.go
waynz0r: the requeue is false by default. i'd rather return blindly otherwise there will be a lot of unnecessary noise in the log for every unsupported mgw. also it is not an error.

waynz0r: maybe we can add a debug level log here.

go.mod
waynz0r: no :)

api/v1alpha1/common.proto
Laci21: I see. Let's not forget to do the first comment here though.

api/v1alpha1/common.proto
Laci21: edited the description

controllers/meshgateway_controller.go
Laci21: I suggested the `requeue: false` bc of this comment here: https://github.com/banzaicloud/istio-operator/pull/713/files/1f902a16213113c8afa64ca641b5287ef2fe7b52#diff-84759ff645b6f0144831537e6dba955fce69a791552223c0626647eb45b9bb65R80-R84 I guess it's not needed there either then, but nvm.

I'd argue with the error log though. There shouldn't be a lot of unsupported MGWs or if there is then the user should be notified about why it is unsupported and probably they'll want to fix the version. If it silently fails or only can be found in debug logs (which are hidden by default), then it'd be hard to find the issue. We logged these kind of errors in the first version of the operator as well and I think it was useful.

Laci21 (approved):

PR #715, opened 2021-08-23
[v2] Istio 1.11 support

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 


Laci21 (commented):
Should we update the `istio.io/client-go` dependency to `1.11` version as well?

(Also, the versions in the sample CRs could be updated as well).

Laci21 (approved):
lgtm, thanks @waynz0r!

PR #769, opened 2021-10-08
[v2] Automate release

Laci21:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

Added a `make release` command to automatically push tags like this:

- `v2.11.0`
- `api/v2.11.0`
- `deploy/charts/v2.0.0`

These GO modules can be imported from other projects.

Based on the tags, CircleCI publishes the following Docker images:

- tag (e.g. `v2.11.0`)
- latest-1.<minor> (e.g. `latest-1.11`)
- latest

CircleCI publishes the Helm chart as well.

### Why?

To be able to release Docker images, Helm charts and go modules can be imported from other projects.

### Additional context

- For the Helm charts we need a tag to trigger a helm chart publish. For the GO embedded chart variable to use a specific chart version we need a tag to be able to use a specific chart version from other GO codes. My suggestion is to use the same tag for these two purposes, e.g. `deploy/charts/v2.0.0`. Placing will trigger a Helm chart publish and can be used right away from other GO codes as well to use the same chart.

- Only `ghcr` images are published, `docker.io` is not used.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,4 +155,6 @@ release: check_release
 	git tag -a ${REL_TAG} -m ${RELEASE_MSG}
 	git tag -a ${API_REL_TAG} -m ${API_RELEASE_MSG}
 	git tag -a ${CHART_REL_TAG} -m ${CHART_RELEASE_MSG}
-	git push origin ${REL_TAG} ${API_REL_TAG} ${CHART_REL_TAG}
+	git push origin ${REL_TAG}
+	git push origin ${API_REL_TAG}
+	git push origin ${CHART_REL_TAG}
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,16 @@
 # Image URL to use all building/pushing image targets
 TAG ?= $(shell git describe --tags --abbrev=0 --match '[0-9].*[0-9].*[0-9]' 2>/dev/null )
-IMG ?= banzaicloud/istio-operator:$(TAG)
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -135,3 +147,12 @@ docker-build: test
...
 	docker push ${IMG}
+
+check_release:
+	@echo "New tags (${REL_TAG}, ${API_REL_TAG} and ${CHART_REL_TAG}) will be pushed to Github, a new Docker image (${REL_TAG}) will be released, and a new Helm chart (${CHART_REL_TAG}) will be released. Are you sure? [y/N] " && read -r ans && [ "$${ans:-N}" = y ]
+
+release: check_release
+	git tag -a ${REL_TAG} -m ${RELEASE_MSG}
+	git tag -a ${API_REL_TAG} -m ${API_RELEASE_MSG}
+	git tag -a ${CHART_REL_TAG} -m ${CHART_RELEASE_MSG}
+	git push origin ${REL_TAG} ${API_REL_TAG} ${CHART_REL_TAG}
```
turip: I would do that as three seperate pushes as CircleCI does not detect all tags if multiple tags are pushed. (It might work for 3)

waynz0r (approved):

asdwsda (approved):

PR #773, opened 2021-10-14
[v2] various fixes

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 


api/v1alpha1/istiocontrolplane_types.go
```diff
@@ -172,6 +173,34 @@ func (icp *PeerIstioControlPlane) GetSpec() *IstioControlPlaneSpec {
...
+
+	for k, v := range rr.Limits {
+		r.Limits[string(k)] = &Quantity{
+			Quantity: v,
+		}
+	}
+
+	return r
+}
+
```
Laci21: Is it deliberate that only the `Limits` are transformed, but the `Requests` are not?

controllers/istiocontrolplane_controller.go
```diff
@@ -892,6 +896,28 @@ func (r *IstioControlPlaneReconciler) reconcileClusterReaderSecret(ctx context.C
...
+	}
+
+	for _, imgw := range imgws.Items {
+		imgw := imgw
+		if imgw.GetSpec().GetIstioControlPlane().Name != icp.Name || imgw.GetSpec().GetIstioControlPlane().Namespace != icp.Namespace {
+			continue
+		}
+		err = util.RemoveFinalizer(ctx, r.Client, &imgw, istioMeshGatewayFinalizerID, false)
+		if err != nil {
+			return err
```
Laci21:
```suggestion
			return errors.WrapIf(err, "could not delete finalizer from istio mesh gateway resource")
```

internal/assets/manifests/istio-discovery/templates/validatingwebhookconfiguration.yaml
```diff
@@ -2,11 +2,7 @@
 apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
-{{- if eq .Values.global.distribution "cisco" }}
   name: {{ include "name-with-namespaced-revision" ( dict "name" "istio-validator" "context" $) }}
-{{- else }}
-  name: istio-validator{{- if not (eq .Values.revision "") }}-{{ .Values.revision }}.{{ .Release.Namespace }}-{{ .Release.Namespace }}{{- end }}
```
Laci21: Why is this not needed anymore for upstream?

Laci21 (approved):

internal/assets/manifests/istio-discovery/templates/validatingwebhookconfiguration.yaml
waynz0r: There is a `VALIDATION_WEBHOOK_CONFIG_NAME` env var that seems to be handled properly in upstream istio as well.

Laci21: I see, thx.

Laci21 (approved):

internal/assets/manifests/istio-discovery/templates/validatingwebhookconfiguration.yaml
waynz0r: scratch that, it was good in 1.10, not handled properly in 1.11, so i reverted the change.

PR #774, opened 2021-10-15
[v2] Update PodDisruptionBudget group version

Laci21:
### Why?

To avoid the following log lines:
```
{"level":"info","ts":"2021-10-15T12:24:09.995Z","logger":"KubeAPIWarningLogger","msg":"policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget"}
```


Laci21:
It would only work on k8s 1.21+, but we support k8s 1.19 and 1.20 as well, so we can only have this later.

PR #783, opened 2021-11-04
[v2] fix permissions in chart and manifests

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### DIFF
--- a/deploy/charts/istio-operator/templates/operator-deployment.yaml
+++ b/deploy/charts/istio-operator/templates/operator-deployment.yaml
@@ -65,7 +65,7 @@ spec:
           {{- if .Values.clusterRegistry.clusterAPI.enabled }}
           - "--cluster-registry-api-enabled"
           {{- end }}
-          {{- if .Values.clusterRegistry.resourceSyncRules.enabled }}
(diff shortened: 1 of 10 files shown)


deploy/charts/istio-operator/templates/operator-deployment.yaml
```diff
@@ -65,7 +65,7 @@ spec:
           {{- if .Values.clusterRegistry.clusterAPI.enabled }}
           - "--cluster-registry-api-enabled"
           {{- end }}
-          {{- if .Values.clusterRegistry.resourceSyncRules.enabled }}
+          {{- if and .Values.clusterRegistry.clusterAPI.enabled .Values.clusterRegistry.resourceSyncRules.enabled }}
```
Laci21, line 68: I deliberately separated this two config options. Most probably it will never happen, but in theory it might be that the user does not want to depend on the cluster name from the cluster CR for some reason, but uses the cluster registry controller anyway and wants to create the ResourceSyncRules.

Anyway, since I see close to zero chance on the scenario I described, I'm okay with this change, just this was my thinking.

Laci21 (approved):

PR #784, opened 2021-11-04
[v2] support conditional crd update 

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

Add support for conditional Istio CRD update based on `managed by` and `resource version (semver style)` labels.

### Why?

The CRDs are managed during ICP reconciliation. As there could be multiple ICP resources the CRD updates must be happen to prevent the multiple reconciliation flows to step on each others toes.

Also there could be situations were the Istio CRDs are already present and managed by different means, so the operator should only update in case the CRDs are managed by it.


hack/crd-labeling-yq.yaml
```diff
@@ -0,0 +1,4 @@
+- command: update
+  path: metadata.labels
+  value:
+    resource.alpha.banzaicloud.io/revision: {{ISTIO_VERSION}}
```
Laci21, line 4: maybe cisco instead of banzai?

internal/components/base/reconcile.go
```diff
@@ -86,17 +97,55 @@ func (rec *Component) ReleaseData(object runtime.Object) (*templatereconciler.Re
...
+					return nil
+				},
+				BeforeUpdateFunc: func(current, desired runtime.Object) error {
+					if o, ok := desired.(client.Object); ok {
+						annotations := o.GetAnnotations()
+						delete(annotations, types.BanzaiCloudManagedComponent)
+						delete(annotations, types.BanzaiCloudRelatedTo)
+						o.SetAnnotations(annotations)
+						k8sutil.SetResourceRevisionLabel(o, rec.supportedIstioVersion)
+						k8sutil.SetManagedByLabel(o, managedByValue)
```
Laci21: could be a function since seems to be the same as in `BeforeCreateFunc`

internal/components/base/reconcile.go
```diff
@@ -86,17 +97,55 @@ func (rec *Component) ReleaseData(object runtime.Object) (*templatereconciler.Re
...
 					}
 
 					return nil
 				},
+				ShouldUpdateFunc: func(current, desired runtime.Object) (bool, error) {
```
Laci21, line 124: how about checking if `current` and `desired` differ from each other to omit unnecessary updates?

pkg/k8sutil/managedbylabels.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+func SetManagedByLabel(obj client.Object, owner string) {
+	obj.SetLabels(util.MergeStringMaps(obj.GetLabels(), map[string]string{
+		types.ManagedByLabel: owner,
+	}))
+}
+
+func GetManagedByLabel(obj client.Object) string {
+	return obj.GetLabels()[types.ManagedByLabel]
```
Laci21, line 33: what if obj.GetLabels() is nil?

pkg/k8sutil/managedbylabels.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/banzaicloud/istio-operator/v2/pkg/util"
+	"github.com/banzaicloud/operator-tools/pkg/types"
+)
+
+func SetManagedByLabel(obj client.Object, owner string) {
+	obj.SetLabels(util.MergeStringMaps(obj.GetLabels(), map[string]string{
```
Laci21, line 27: what if obj.GetLabels() is nil?

pkg/k8sutil/resourceversion.go
```diff
@@ -0,0 +1,60 @@
...
+		labels = make(map[string]string)
+	}
+
+	obj.SetLabels(util.MergeStringMaps(labels, map[string]string{
+		resourceRevisionLabel: revision,
+	}))
+}
+
+func GetResourceRevisionLabel(obj client.Object) string {
+	return obj.GetLabels()[resourceRevisionLabel]
```
Laci21, line 43: what if obj.GetLabels() is nil?

Laci21 (approved):

internal/components/base/reconcile.go
waynz0r: that is handled by the ot reconciler

pkg/k8sutil/managedbylabels.go
waynz0r: it is handled by golang

pkg/k8sutil/managedbylabels.go
waynz0r: it is handled within util.MergeStringMaps

pkg/k8sutil/resourceversion.go
waynz0r: it is handled by golang

Laci21 (approved):

PR #800, opened 2021-12-07
Fix label selector to list the Istio root ca configmaps

Laci21:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

- Pass all filters to controller-runtime client `.List()` function as `client.MatchingLabelsSelector`s
- Add function comment and change logging as well

### Why?

- In the original solution, the latter `client.MatchingLabels` `ListOption` overrode the previous `client.MatchingLabelsSelector` `ListOption`. That's why a single `client.MatchingLabelsSelector` is passed instead to be able to filter for all conditions.

### Checklist

- [x] Implementation tested

### DIFF
--- a/controllers/istiocontrolplane_controller.go
+++ b/controllers/istiocontrolplane_controller.go
@@ -307,8 +307,7 @@ func (r *IstioControlPlaneReconciler) reconcile(ctx context.Context, icp *servic
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


waynz0r (approved):
thanks @Laci21!

PR #803, opened 2021-12-12
support additional external service for istio mesh gateways

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

Adds support to have an additional k8s headless service resource with name is suffixed with `-external` for istio mesh gateways configured as load balancer type. The additional service has static endpoints for the external ip address(es) of the mesh gateway.

### Why?

The additional service is added to support use-cases where easy access is needed for the external address of any mesh gateway. The feature is automatically turned on for the mesh expansion gateway.

The feature is turned on with `meshgateway.istio.servicemesh.cisco.com/generate-external-service annotation on the imgw resource set to true.

### DIFF
--- a/api/v1alpha1/istiomeshgateway_types.go
+++ b/api/v1alpha1/istiomeshgateway_types.go
@@ -79,12 +79,13 @@ func (p *IstioMeshGatewayWithProperties) SetDefaults() {
 }
 
 type IstioMeshGatewayProperties struct {
-	Revision              string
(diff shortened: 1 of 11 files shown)


Laci21 (approved):
Thank you, @waynz0r!

PR #805, opened 2021-12-20
Generate OpenAPI descriptor

gallotamas:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?
Generate OpenAPI descriptor for istio operator

### Why?
External tools (e.g. code editor) might need open api schemas so let's generate it if we already have the toolset for it.

### Additional context
perFile openAPI schema generation doesn't work but it works by generating all types into a single file.
We might want to fix the perFile schema generation (I couldn't make it work so far) if the API becomes too big.


waynz0r (approved):

api/options/options.proto
```diff
@@ -0,0 +1,29 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+import "google/protobuf/descriptor.proto";
+
+package istio_operator.v2.api.options;
+
+option go_package = "github.com/banzaicloud/istio-operator/api/v2/options;options";
```
Laci21: just checking: `options;options` is intended to be like this?

Laci21 (approved):

api/options/options.proto
waynz0r: it was intended, but `;options` is not needed, so it got removed.

PR #815, opened 2022-01-26
Fix #814

### LINKED ISSUE #814: Istio MeshGateway Deployment lacking status port exposal.
**Describe the bug**
Istio MeshGateway Deployment container does not expose status port (15020).

**Steps to reproduce the issue:**
Deploy a ControlPlane. Deploy a MeshGateway for the CP.

**Expected behavior**
No errors regarding scraping port.

Tanemahuta:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #814

### What's in this PR?
exposal of status port of istio proxy.

### Checklist

- [x] Implementation tested

### DIFF
--- a/internal/assets/manifests/istio-meshgateway/templates/deployment.yaml
+++ b/internal/assets/manifests/istio-meshgateway/templates/deployment.yaml
@@ -50,10 +50,10 @@ spec:
             {{- end }}
             - containerPort: 15020
               protocol: TCP
-              name: http-merged-prom
+              name: http-int-dbg
             - containerPort: 15021
               protocol: TCP
-              name: http-health-check
+              name: http-ext-health
             - containerPort: 15090
               protocol: TCP
               name: http-envoy-prom
--- a/internal/assets/manifests/istio-meshgateway/templates/deployment.yaml
+++ b/internal/assets/manifests/istio-meshgateway/templates/deployment.yaml
@@ -48,12 +48,15 @@ spec:
             - containerPort: {{ $val.targetPort | default $val.port }}
               protocol: {{ $val.protocol | default "TCP" }}
             {{- end }}
+            - containerPort: 15020
+              protocol: TCP
+              name: http-merged-prom
+            - containerPort: 15021
+              protocol: TCP
+              name: http-health-check
             - containerPort: 15090
               protocol: TCP
               name: http-envoy-prom
-            - containerPort: 15020
-              protocol: TCP
-              name: status
         {{- if not $gateway.runAsRoot }}
           securityContext:
             allowPrivilegeEscalation: false
--- a/internal/assets/manifests/istio-meshgateway/templates/deployment.yaml
+++ b/internal/assets/manifests/istio-meshgateway/templates/deployment.yaml
@@ -51,6 +51,9 @@ spec:
             - containerPort: 15090
               protocol: TCP
               name: http-envoy-prom
+            - containerPort: 15020
+              protocol: TCP
+              name: status
         {{- if not $gateway.runAsRoot }}
           securityContext:
             allowPrivilegeEscalation: false
--- a/internal/components/istiomeshgateway/testdata/imgw-expected-resource-dump.yaml
+++ b/internal/components/istiomeshgateway/testdata/imgw-expected-resource-dump.yaml
@@ -232,6 +232,9 @@ spec:
         - containerPort: 15090
           name: http-envoy-prom
           protocol: TCP
+        - containerPort: 15020
+          name: status
+          protocol: TCP
         resources:
           limits:
             cpu: "2"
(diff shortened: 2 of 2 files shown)


internal/assets/manifests/istio-meshgateway/templates/deployment.yaml
```diff
@@ -51,6 +51,9 @@ spec:
             - containerPort: 15090
               protocol: TCP
               name: http-envoy-prom
+            - containerPort: 15020
+              protocol: TCP
+              name: status
```
Laci21: Can we rename this to have `http-` prefix to align with Istio port naming conventions?
```suggestion
              name: http-status
```

Laci21:
There are issues with running the CI for forks right now, so disregard that for now please.

Regarding the `Mergeable` failure, pls remove the `To Do` section from the PR description.

Also, pls rebase the branch.

internal/assets/manifests/istio-meshgateway/templates/deployment.yaml
Tanemahuta: Yes, you are correct, I renamed it to the health check port and I also added the combined metrics port to be exposed in the next commit (ref: https://istio.io/latest/docs/ops/deployment/requirements/#ports-used-by-istio).

internal/assets/manifests/istio-meshgateway/templates/deployment.yaml
```diff
@@ -48,6 +48,12 @@ spec:
             - containerPort: {{ $val.targetPort | default $val.port }}
               protocol: {{ $val.protocol | default "TCP" }}
             {{- end }}
+            - containerPort: 15020
+              protocol: TCP
+              name: http-merged-prom
+            - containerPort: 15021
+              protocol: TCP
+              name: http-health-check
```
Laci21: If the user would like to expose port 15021 publicly, so on the `service` resource as well, then they should add this port to their `IstioMeshGateway` resource spec. If they do that, the 15021 will be appended to the `containerPorts` as well, as you can see a few lines above (around 48) in this file. But if we add port 15021 explicitly as you propose in this PR, then it would be duplicated. That's why I suggest **not to add 15021 here**, only 15020 as it is most likely not intended to be exposed publicly (on the `service`).

Tanemahuta: Sorry, I might be wrong here, but exposing it on the service would IMHO need an explicit
```
  ports:
  [...]
    - protocol: TCP
      port: 15021
      targetPort: 15021
```
to work, wouldn't it?

Laci21: Yes,  this should be added to the `IstioMeshGateway` spec if you'd like to expose this port on the service.

Tanemahuta: I did not mean to expose the port to the outside of the namespace network. But since this is a health check port, it might be really handy to expose the container port.

waynz0r:
I would suggest the following change:

```yaml
          ports:
            {{- $defaultPorts := list 15020 15021 15090 }}
            {{- range $key, $val := $service.ports }}
            {{- $port := (($val.targetPort | default $val.port) | int) }}
            {{- if not (has $port $defaultPorts) }}
            - containerPort: {{ $port }}
              protocol: {{ $val.protocol | default "TCP" }}
            {{- end }}
            {{- end }}
            - containerPort: 15020
              protocol: TCP
              name: http-int-dbg
            - containerPort: 15021
              protocol: TCP
              name: http-ext-hlth
            - containerPort: 15090
              protocol: TCP
              name: http-envoy-prom
```

The port names were changed because there is a 15 char limit on the container port names.

Also, just for clarity, the container ports within the pod spec is only informational purposes and has no effect whether a port is exposed or not.

Tanemahuta:
> I would suggest the following change:
> 
> ```yaml
>           ports:
>             {{- $defaultPorts := list 15020 15021 15090 }}
>             {{- range $key, $val := $service.ports }}
>             {{- $port := (($val.targetPort | default $val.port) | int) }}
>             {{- if not (has $port $defaultPorts) }}
>             - containerPort: {{ $port }}
>               protocol: {{ $val.protocol | default "TCP" }}
>             {{- end }}
>             {{- end }}
>             - containerPort: 15020
>               protocol: TCP
>               name: http-int-dbg
>             - containerPort: 15021
>               protocol: TCP
>               name: http-ext-hlth
>             - containerPort: 15090
>               protocol: TCP
>               name: http-envoy-prom
> ```
> 
> The port names were changed because there is a 15 char limit on the container port names.
> 
> Also, just for clarity, the container ports within the pod spec is only informational purposes and has no effect whether a port is exposed or not.

yip, I know. one can leave them out, but I have the opinion they should be in there (at least for documentation purposes => self describing resource).

waynz0r (approved):

PR #838, opened 2022-02-22
[v2] add custom patch maker for proto based api

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

Adds a custom patch maker for k8s object matching library to support Istio proto based api/resources.

### Why?

There is an issue using the default patch maker in the k8s object matcher library with proto based Istio api/resources. This is mainly, because there is a discrepancy between the field names in `json` and `proto` tags for some fields and that causes issue with the upstream k8s strategic merge patcher library field detection.

### Checklist

- [x] Implementation tested


pkg/util/patch.go
```diff
@@ -0,0 +1,146 @@
...
+		return nil, err
+	}
+
+	return strategicpatch.CreateTwoWayMergePatchUsingLookupPatchMeta(original, modified, schema, p.PreconditionFuncs...)
+}
+
+func (p *ProtoCompatibleK8sStrategicMergePatcher) CreateThreeWayMergePatch(original, modified, current []byte, dataStruct interface{}) ([]byte, error) {
+	lookupPatchMeta, err := NewPatchMetaFromStruct(dataStruct)
+	if err != nil {
+		return nil, errors.WrapWithDetails(err, "Failed to lookup patch meta", "current object", dataStruct)
```
Laci21:
```suggestion
		return nil, errors.WrapWithDetails(err, "failed to lookup patch meta", "current object", dataStruct)
```

Laci21 (approved):
lgtm, thank you very much, @waynz0r!

PR #853, opened 2022-06-09
fix ICP replicas reconciliation

luciferinlove:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?
Fixed a bug where the istiod number of replicas couldn't be changed in the IstioControlPlane resource.


kristofgyuracz (approved):
LGTM

waynz0r (changes requested):
@LuciferInLove,

in its current state this doesn't really solves the base problem. Although the replica count is set for the deployment it will be overridden by the HPA since the autoscale is still enabled by default.

After internal discussion the expected behaviour for the CR fields are the following:

- there is a 1-5 autoscale enabled by default
- if only the `count` is set then the autoscale should be disabled and the count is set for the deployment
- `min` or `max` has precedence, if either one of those  is set (regardless of the value of the `count` if any) then those override the default setting and the autoscale is enabled accordingly
- the actual count always >= `min` and is always set for the deployment

```
# autoscale disabled, fixed replica count is set for the deployment
replicas:
  count: 4
```

```
# autoscale enabled (2-5), count = 2
replicas:
  min: 2
```

```
# autoscale enabled (2-4), count = 3
replicas:
  count: 3
  min: 2
  max: 4
```

waynz0r (approved):

PR #860, opened 2022-08-15
Istio1.15 support

luciferinlove:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | yes
| Deprecations?   | no
| Related tickets |

### What's in this PR?

Istio1.15 support

### Checklist

- [X] Implementation tested
- [X] User guide and development docs updated (if needed)


api/v1alpha1/common.proto
```diff
@@ -60,15 +57,15 @@ message ContainerImageConfiguration {
 
     // ImagePullSecrets is an optional list of references to secrets to use for pulling any of the images.
     // +optional
-    repeated k8s.io.api.core.v1.LocalObjectReference imagePullSecrets = 4 [(gogoproto.nullable) = false];
```
Laci21, line 63: Do we have the same functionality as `[(gogoproto.nullable) = false]`?

Laci21: Or for `[(gogoproto.wktpointer) = true]` or others?

api/v1alpha1/common.proto
```diff
@@ -125,24 +122,24 @@ message BaseKubernetesResourceConfig {
     // +optional
     string priorityClassName = 10;
 
-    // google.protobuf.Int32Value replicaCount = 1 [(gogoproto.wktpointer) = true];
+    // google.protobuf.Int32Value replicaCount = 1;
```
Laci21: Not sure why this is commented out. If not needed we could delete this, otherwise we should use this field. I guess we don't need it, but did not check.

api/v1alpha1/istiocontrolplane.proto
```diff
@@ -126,7 +125,7 @@ message IstioControlPlaneSpec {
 }
 
 enum ModeType {
-    UNSPECIFIED = 0;
+    ModeType_UNSPECIFIED = 0;
```
Laci21, line 128: Is this still not working if you use the same name `UNSPECIFIED` multiple places? Other option, what you simply delete this field, would that work maybe?

api/v1alpha1/istiocontrolplane_types.go
```diff
@@ -54,16 +53,20 @@ type IstioControlPlane struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   *IstioControlPlaneSpec  `json:"spec,omitempty"`
-	Status IstioControlPlaneStatus `json:"status,omitempty"`
+	Spec   *IstioControlPlaneSpec   `json:"spec,omitempty"`
+	Status *IstioControlPlaneStatus `json:"status,omitempty"`
```
Laci21, line 57: Why was this needed exactly?

deploy/charts/istio-operator/Chart.yaml
```diff
@@ -9,7 +9,7 @@ sources:
 icon: https://istio.io/latest/img/istio-whitelogo-bluebackground-framed.svg
 
 # Based on support status of Istio releases: https://istio.io/latest/docs/releases/supported-releases/#support-status-of-istio-releases
-kubeVersion: ">= 1.20.0-0 <= 1.23.0-0"
+kubeVersion: ">= 1.21.0-0 <= 1.24.3-0"
```
Laci21: this will probably be >= 1.22.0-0 <= 1.26.0-0

deploy/charts/istio-operator/Chart.yaml
```diff
@@ -9,7 +9,7 @@ sources:
...
 
-version: 2.0.12
-appVersion: "v2.13.4"
+version: 2.0.13
+appVersion: "v2.14.0"
```
Laci21: We usually release rc releases first for testing, but I'm not consisting if you can test otherwise as well easily, let's just not release the stable version until the code is reviewed/solution is tested.

deploy/charts/istio-operator/README.md
```diff
@@ -1,11 +1,11 @@
 # Istio-operator chart
 
-[Istio-operator](https://github.com/banzaicloud/istio-operator/tree/release-1.13) is a Kubernetes operator to deploy and manage [Istio](https://istio.io/) resources for a Kubernetes cluster.
+[Istio-operator](https://github.com/banzaicloud/istio-operator/tree/release-1.15) is a Kubernetes operator to deploy and manage [Istio](https://istio.io/) resources for a Kubernetes cluster.
 
 ## Prerequisites
 
 - Helm3
-- Kubernetes 1.19.0 - 1.22.x
+- Kubernetes 1.21.0 - 1.24.x
```
Laci21, line 8: same as above

deploy/charts/istio-operator/templates/operator-rbac.yaml
```diff
@@ -365,6 +365,15 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
```
Laci21, line 368: this might need to be added to the controller.go file as well to generate it for the kustomize manifests

internal/assets/manifests/istio-discovery/templates/telemetryv2_1.12.yaml
```diff
@@ -14,6 +14,7 @@ metadata:
   labels:
```
Laci21, line 14: I think it's fine to remove these 1.12 CRs this time.

internal/components/sidecarinjector/testdata/icp-expected-values.yaml
```diff
@@ -1,6 +1,6 @@
-revision: cp-v113x
+revision: cp-v115x
 deployment:
-  image: banzaicloud/istio-sidecar-injector:v1.10.4-bzc.1
+  image: banzaicloud/istio-sidecar-injector:v1.15.0-bzc.1
```
Laci21, line 3: Does this image actually exist?

internal/components/sidecarinjector/testdata/icp-test-cr.yaml
```diff
@@ -30,7 +30,7 @@ spec:
         type: RollingUpdate
         rollingUpdate:
           maxUnavailable: 1
-      image: banzaicloud/istio-sidecar-injector:v1.10.4-bzc.1
+      image: banzaicloud/istio-sidecar-injector:v1.15.0-bzc.1
```
Laci21, line 33: same Q here

internal/util/template.go
```diff
@@ -106,6 +106,11 @@ func valueIfTemplateFunc(value interface{}) (string, error) {
 			return "", nil
 		}
 
+		m, ok := reflect.TypeOf(value).MethodByName("GetValue")
+		if ok {
+			value = m.Func.Call([]reflect.Value{reflect.ValueOf(value)})[0].Interface()
```
Laci21, line 111: Could you pls explain to me what's happening here and why exactly? I'd like to learn :)

Laci21 (commented):
Great work, @LuciferInLove! I've just added some nits/questions.

api/v1alpha1/istiocontrolplane.proto
luciferinlove: New buf doesn't support fields defined in multiple places. There was a recommendation to rename the fields in such a manner.

api/v1alpha1/istiocontrolplane_types.go
luciferinlove: Because after moving to golang/protobuf this field contains mutex, so there were copylocks.

internal/components/sidecarinjector/testdata/icp-expected-values.yaml
luciferinlove: No, not yet

internal/util/template.go
luciferinlove: If a structure has `GetValue` method - call this method to get a value. This is needed because after moving to golang/protobuf there are wrappers for types:
https://github.com/golang/protobuf/blob/master/ptypes/wrappers/wrappers.pb.go

The same is for methods in `.tpl` files.

api/v1alpha1/common.proto
Laci21: Talked about this offline, should be fine.

internal/components/sidecarinjector/testdata/icp-expected-values.yaml
Laci21: And will it? Seems like we haven't released a new one since `1.10.4`, will we have a new one for some reason?

internal/util/template.go
Laci21: Thanks!

Laci21 (approved):

internal/components/sidecarinjector/testdata/icp-expected-values.yaml
luciferinlove: This is only testdata, no real images will be downloaded.

Laci21: oh, okay :)

PR #866, opened 2022-09-01
add permissions for roles to support openshift policies

kristofgyuracz:
### What's in this PR?
Added permissions to the manager role to be able to set finalizers on operator-managed CRs.

### Why?
Openshift does not permit this by default.


kristofgyuracz:
docs in progress.

PR #869, opened 2022-10-12
Updated cluster_secret.go to support k8s 1.24

Kuvesz:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
So when using cluster-registry I encountered an error where the IstioControlPlane had a message saying "could not find secret reference for sa" which is an error I encountered before and happens on K8S 1.24.

### Why?
The reason why this happens is detailed [here](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#no-really-you-must-read-this-before-you-upgrade-1), the fix I did was taken straight from the cluster-registry controller repo [here](https://github.com/cisco-open/cluster-registry-controller/blob/master/pkg/util/cluster_secret.go)

### Checklist

- [X] Implementation tested

### DIFF
--- a/pkg/k8sutil/cluster_secret.go
+++ b/pkg/k8sutil/cluster_secret.go
@@ -17,7 +17,6 @@ limitations under the License.
 package k8sutil
 
 import (
-	"bytes"
 	"context"
 	"net"
 	"net/url"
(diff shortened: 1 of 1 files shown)


Laci21 (approved):
thank you, @Kuvesz!

pregnor (approved):

PR #870, opened 2022-10-12
Bump versions for new tags

Kuvesz:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Forgot to bump up the versions so I can create new tags. Sorry about he double PR.

### Why?
Same as above.

### Checklist

- [X] Implementation tested


pregnor (commented):
LGTM

deploy/charts/istio-operator/Chart.yaml
```diff
@@ -11,5 +11,5 @@ icon: https://istio.io/latest/img/istio-whitelogo-bluebackground-framed.svg
 # Based on support status of Istio releases: https://istio.io/latest/docs/releases/supported-releases/#support-status-of-istio-releases
 kubeVersion: ">= 1.21.0-0 < 1.26.0-0"
 
-version: 2.0.16
-appVersion: "v2.14.0"
+version: 2.0.17
+appVersion: "v2.16.0"
```
Laci21:
```suggestion
appVersion: "v2.15.1"
```

We have an unwritten convention that the appVersion's minor version should be the same as Istio's minor version. E.g. for Istio 1.15, we should have `v2.15.x` appVersions. (We already have `v2.15.0` released, that's why I'm suggesting `v2.15.1`.)

Also, we want to write down this convention to the docs, sorry about this.

(Btw I see these versions tagged already `v2.16.0`, `deploy/charts/v2.0.17`, `chart/istio-operator/2.0.17` and `api/v2.16.0`, even though this PR is not merged yet. I don't know why that's the case, but they shouldn't be tagged like that without merging the corresponsing PR.)

Laci21 (changes requested):

deploy/charts/istio-operator/Chart.yaml
pregnor: IMO we should document this

Laci21: Yes, agreed. - "Also, we want to write down this convention to the docs, sorry about this."

Kuvesz: Okay, I'll change it. Also I just realized that I forgot these when I created the tags, I'll get rid of those

Laci21 (approved):

bartam1 (approved):

Issue #857, opened 2022-08-12
Ability to configure modifications to default injection templates


tiswanso:
**Is your feature request related to a problem? Please describe.**
The istio-operator's built in injection templates are not able to be modified without [applying annotations](https://istio.io/latest/docs/reference/config/annotations/) to each application deployment spec's pod template.  For istio-admins to be able to add custom settings in the injection templates for non-annotated workloads, a mechanism is needed to customize the default injection templates.

**Describe the solution you'd like to see**
The current istio community operator has the following mechanism to merge custom settings with the default sidecar injection template https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#custom-templates-experimental

The following is the configuration interface for the istio.io operator to specify what to merge into the sidecar injection template:
```
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
metadata:
  name: istio
spec:
  values:
    sidecarInjectorWebhook:
      templates:
        custom: |
          spec:
            containers:
            - name: istio-proxy
              env:
              - name: GREETING
                value: hello-world
```

Still the `templates` annotation at the app deployment/pod level is needed to specify the merge:
> For example, to apply the default template and our customization, you can set inject.istio.io/templates=sidecar,custom

It'd be a better admin experience to allow a similar mechanism to actually specify customizations to the default injection templates.  Possibly the config interface would be better if it followed a similar pattern as the istio pod [annotations](https://istio.io/latest/docs/reference/config/annotations/), e.g. something like:

```
apiVersion: servicemesh.cisco.com/v1alpha1
  kind: IstioControlPlane
  metadata:
   ...
  spec:
     sidecarInjector:
        templateCustomization:
          sidecar:
             userVolume:
                 - name: custom-id
                    secret:
                       optional: true
                       {{- if (isset .ObjectMeta.Annotations `sidecar.istio.io/istioCustomIdSecretName`) }}
                       secretName: "{{ index .ObjectMeta.Annotations `sidecar.istio.io/istioCustomIdSecretName` }}"
                       {{- else }}
                       secretName: placeholder
                       {{- end }}
            userVolumeMount:
                 - name: custom-id
                    mountPath: /etc/istio/custom-id
                    readOnly: true
          gateway:
             proxyImage:  ghcr.io/myorg/special-gateway-proxy:vX.y.z
```

[...]

PR #872, opened 2022-10-19
Ability to configure modifications to default injection templates

kristofgyuracz:
| Q               | A
| --------------- | ---
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | #857 

### What's in this PR?
Add ability to - instead of replacing the whole default sidecar/gateway injector templates - patch/overwrite default injector templates.


api/v1alpha1/istiocontrolplane.proto
```diff
@@ -135,6 +135,17 @@ message SidecarInjectorConfiguration {
     BaseKubernetesResourceConfig deployment = 1;
     // Service spec
     Service service = 2;
+    // Template spec
```
Laci21: nit:
```suggestion
    // Sidecar injection templates spec
```

Also, we could probably write a few sentences that with this field the default `sidecar` and/or `gateway` injection template can be overriden and also custom templates can be defined, which then can be used on pods with the `inject.istio.io/templates=custom` annotations.

api/v1alpha1/istiocontrolplane.proto
```diff
@@ -135,6 +135,17 @@ message SidecarInjectorConfiguration {
...
+    Templates templates = 3;
```
Laci21: optional nit, just to clarify templates:
```suggestion
    SidecarInjectionTemplates templates = 3;
```

api/v1alpha1/istiocontrolplane.proto
```diff
@@ -135,6 +135,17 @@ message SidecarInjectorConfiguration {
...
+}
+
+message Templates {
+    // Overrides for sidecar templates (will be mapped as sidecarOverrides)
```
Laci21: Since these are user facing specs, I'd probably suggest to not mention the `sidecarOverrides` as it is only an implementation detail.
```suggestion
    // Overrides for the default "sidecar" injection template
    // This spec will be merged into the default "sidecar" template
```

api/v1alpha1/istiocontrolplane.proto
```diff
@@ -135,6 +135,17 @@ message SidecarInjectorConfiguration {
...
+    string sidecar = 1;
+    // Overrides for gateway templates (will be mapped as gatewayOverrides)
```
Laci21: same as above

api/v1alpha1/istiocontrolplane.proto
```diff
@@ -135,6 +135,17 @@ message SidecarInjectorConfiguration {
...
+    string gateway = 2;
+    // Custom templates
+    string custom = 3;
```
Laci21: If I understand correctly here the user will be able to define a custom template called `custom`. With this name only and cannot specify other templates.

I'd rather suggest to have a solution where the user can define multiple custom templates with any arbitrary name.

Implementation-wise I imagine this should be either a <string,string> map (where key is the name of the template and value is the template string itself) or a list of objects (where the object has two field, a string name of the template and a string template itself).

Map:
```yaml
customTemplates:
  custom1: |
          spec:
            containers:
            - name: foo
              image: bar 
  custom2: |
          spec:
            containers:
            - name: foo
              image: bar
```

List:
```yaml
customTemplates:
  - name: custom1
    template: |
          spec:
            containers:
            - name: foo
              image: bar 
  - name: custom2
    template: |
          spec:
            containers:
            - name: foo
              image: bar
```

At first, I'm thinking the map based solution is enough, but there can be counter arguments. Feel free to propose what you think makes most sense (maybe a third solution).

controllers/meshgateway_controller.go
```diff
@@ -126,7 +126,7 @@ func (r *IstioMeshGatewayReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		return istiomeshgateway.NewChartReconciler(helmReconciler, servicemeshv1alpha1.IstioMeshGatewayProperties{
 			Revision:                fmt.Sprintf("%s.%s", icp.GetName(), icp.GetNamespace()),
 			EnablePrometheusMerge:   utils.BoolPointer(enablePrometheusMerge),
-			InjectionTemplate:       "gateway",
+			InjectionTemplate:       "gateway, gatewayOverrides",
```
Laci21: Not very important, but I'm thinking if it would be better that in case there is no `gatewayOverrides` (or `sidecarOverrides`) provided through the ICP spec, then it might be more optimal to have only `gateway` (or `sidecar`) as the injection template as usual.

For a few reasons:
- I know it's an empty spec merge, but still no need to do that template merge on the Istio side for the overrides
- It might be confusing to the user why that overrides is there if it's specified nowhere by the user in a normal installation (without providing overrides in the ICP spec)
- Also, if there is a bug in our implementation at any time in the future this solution won't mess up default installations without these overrides

What do you think?

internal/assets/manifests/istio-discovery/values.yaml.tpl
```diff
@@ -107,11 +107,29 @@ pilot:
 {{ $x | indent 2 }}
 {{- end }}
 
-{{- if .GetSpec.GetHttpProxyEnvs }}
```
Laci21, line 110: Maybe I'm missing sth but just to double check is this deliberately deleted?

internal/assets/manifests/istio-discovery/values.yaml.tpl
```diff
@@ -107,11 +107,29 @@ pilot:
...
 sidecarInjectorWebhook:
   # Supported only in Cisco provided istio-proxy images
 {{ toYamlIf (dict "value" .GetSpec.GetHttpProxyEnvs "key" "httpProxyEnvs") | indent 2 }}
+  templates:
+{{- if .GetSpec.GetSidecarInjector.GetTemplates.GetSidecar }}
+{{ valueIf (dict "key" "sidecarOverrides" "value" .GetSpec.GetSidecarInjector.GetTemplates.GetSidecar) | indent 4 }}
 {{- end }}
+{{- if not .GetSpec.GetSidecarInjector.GetTemplates.GetSidecar }}
```
Laci21: maybe an `else` instead of `if not`?

internal/assets/manifests/istio-discovery/values.yaml.tpl
```diff
@@ -107,11 +107,29 @@ pilot:
...
+    sidecarOverrides: ""
+{{- end }}
+{{- if .GetSpec.GetSidecarInjector.GetTemplates.GetGateway }}
+{{ valueIf (dict "key" "gatewayOverrides" "value" .GetSpec.GetSidecarInjector.GetTemplates.GetGateway) | indent 4 }}
+{{- end }}
+{{- if not .GetSpec.GetSidecarInjector.GetTemplates.GetGateway }}
```
Laci21: same as above

internal/assets/manifests/istio-discovery/values.yaml.tpl
```diff
@@ -107,11 +107,29 @@ pilot:
...
+    gatewayOverrides: ""
+{{- end }}
+{{- if .GetSpec.GetSidecarInjector.GetTemplates.GetCustom }}
+{{ valueIf (dict "key" "custom" "value" .GetSpec.GetSidecarInjector.GetTemplates.GetCustom) | indent 4 }}
+{{- end }}
+{{- if not .GetSpec.GetSidecarInjector.GetTemplates.GetCustom }}
```
Laci21: same as above

Laci21 (commented):
Thank you, @kristofgyuracz!

I have a few suggestions. The most important one is probably about how to handle multiple custom templates, the others are less important.

Also, I'll leave it here for further reference that we should add user facing doc for this nice feature.

api/v1alpha1/istiocontrolplane.proto
```diff
@@ -135,6 +135,24 @@ message SidecarInjectorConfiguration {
     BaseKubernetesResourceConfig deployment = 1;
     // Service spec
     Service service = 2;
+    // SidecarInjectionTemplates spec
```
Laci21: Since this is human readable comment, I'd suggest like this:
```suggestion
    // Sidecar injection templates spec
```

Also, we could probably write a few sentences that with this field the default `sidecar` and/or `gateway` injection template can be overriden and also custom templates can be defined, which then can be used on pods with the `inject.istio.io/templates=custom` annotations.

api/v1alpha1/istiocontrolplane.proto
```diff
@@ -135,6 +135,24 @@ message SidecarInjectorConfiguration {
...
+    SidecarInjectionTemplates templates = 3;
+    // Default templates to be injected
+    repeated string defaultSidecarTemplates = 4;
```
Laci21: Probably not needed as discussed offline.

config/manager/kustomization.yaml
```diff
@@ -4,4 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: ghcr.io/banzaicloud/istio-operator
+  newName: kristofgyuracz/istio-operator
+  newTag: 1.1.11
```
Laci21: Please, don't commit this.

internal/assets/manifests/istio-discovery/values.yaml.tpl
Laci21: What about this comment?

internal/assets/manifests/istio-discovery/values.yaml.tpl
Laci21: What about this comment?

internal/assets/manifests/istio-discovery/values.yaml.tpl
Laci21: What about this comment?

internal/assets/manifests/istio-discovery/values.yaml.tpl
```diff
@@ -107,11 +107,34 @@ pilot:
...
+{{- end }}
+{{- if not .GetSpec.GetSidecarInjector.GetTemplates.GetGateway }}
+    gatewayOverrides: ""
+{{- end }}
+{{- if .GetSpec.GetSidecarInjector.GetTemplates.GetCustomTemplates }}
+{{ range $customTemplate := .GetSpec.GetSidecarInjector.GetTemplates.GetCustomTemplates }}
+{{ valueIf (dict "key" $customTemplate.Name "value" $customTemplate.Template) | indent 4 }}
+{{ end }}
+{{- end }}
+{{- if .GetSpec.GetSidecarInjector.GetTemplates.GetSidecar }}
```
Laci21, line 124: this same if is used in line 114. maybe setting the `defaultTemplates` can be done as well?

internal/components/discovery/testdata/icp-test-cr.yaml
```diff
@@ -4,177 +4,41 @@ metadata:
...
+              - name: istio-proxy
+                env:
+                - name: TEMPLATE
+                  value: custom2
   meshID: mesh1
   clusterID: demo-cluster1
   version: "1.15.3"
   mode: ACTIVE
   networkName: network1
-  containerImageConfiguration:
```
Laci21: Is everything below this deliberately deleted?

Laci21: I guess CI broke because of this.

kristofgyuracz: fixed

internal/assets/manifests/istio-discovery/values.yaml.tpl
kristofgyuracz: merged

internal/assets/manifests/istio-discovery/values.yaml.tpl
kristofgyuracz: before this feature, this field was only used in case proxyenvs is set (see next 2 lines). now these values will be set either way.

Laci21: Why exactly will "these values will be set either way"?

Isn't there a matching `end` section that should be deleted in this case as well?

api/v1alpha1/istiocontrolplane.proto
kristofgyuracz: done.

api/v1alpha1/istiocontrolplane.proto
Laci21: I think the comment is not yet updated as suggested.

internal/assets/manifests/istio-discovery/values.yaml.tpl
kristofgyuracz: the matching end is not detected in the diff, see line 114. it is "reused"

Laci21: Ah, I see, thanks!

Laci21 (approved):
LGTM, thank you very much! Nice work, @kristofgyuracz!

PR #910, opened 2023-03-09
[v1.15.x] CNI related fixes

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

CNI component related fixes to properly support multiple CNI daemonsets for each control plane on one cluster.

- [x] Implementation tested

### DIFF
--- a/internal/assets/manifests/istio-cni/templates/daemonset.yaml
+++ b/internal/assets/manifests/istio-cni/templates/daemonset.yaml
@@ -71,6 +71,8 @@ spec:
                   key: cni_network_config
             - name: CNI_NET_DIR
               value: {{ default "/etc/cni/net.d" .Values.cni.cniConfDir }}
+            - name: ISTIO_REVISION
+              value: {{ include "namespaced-revision" . }}
             # Deploy as a standalone CNI plugin or as chained?
             - name: CHAINED_CNI_PLUGIN
               value: "{{ .Values.cni.chained }}"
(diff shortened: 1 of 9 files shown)


Laci21 (approved):
lgtm, thank you very much, @waynz0r!

internal/assets/manifests/istio-cni/templates/daemonset.yaml
```diff
@@ -71,6 +71,8 @@ spec:
                   key: cni_network_config
```
tiswanso, line 71: edit: This comment is for CNI_CONF_NAME env var not being modified in this changeset

Will try to followup more closely on this but I was assuming the filename would need to be `name-with-revision` as well.  In the chat test info you showed I saw the filename had rev appended so probably this works with this changeset so I'll try to grok whether there's any missing conditions.

waynz0r: The CNI_CONF_NAME value is not generated, but it specified by the user in the icp spec, so it is up to the user the set it properly.

waynz0r: Also, it is the actual filename of the config on the disk, it is not the `name` property inside the config.

tiswanso (approved):

